PR #23, opened 2018-02-07
Unit and system tests

mcafaro:
**Testing strategy:**
MatNWB contains two main software interfaces of interest:
1. The interface that generates a MATLAB API for reading and writing NWB files
2. The interface exposed by this generated MATLAB API

I primarily focused on unit testing interface 1 and system testing interface 2.

*Unit testing interface 1:*
For unit testing interface 1, I concentrated on the parseClass and writeClass functions. The parseClass function was tested by passing small sections of the NWB spec language into the parser and verifying the actual output struct against an expected struct. Because there is currently no spec I am aware of that defines the expected output of parseClass, I did my best to define one through examining the expected input of writeClass. With these tests, I tried to cover the full extent of the NWB spec language.

The writeClass function was tested by passing in small structs (equivalent to those produced by parseClass) into the writer and verifying the generated Class file was formed and/or functions in an expected way. Again, because there is currently no spec I am aware of that defines the expected translation from YAML to MATLAB, I did my best to define one through examining the PyNWB API. With these tests, I tried to cover the full extent of functionality supported by MatNWB. In some cases where functionality is not yet supported, I created placeholder tests that are currently ignored (filtered) in the test suite.  

*System testing interface 2:*
For system testing interface 2, I concentrated on two test types using the generated API produced by the core NWB format spec:
1. Round-trip from MatNWB to disk and back again
2. Out from MatNWB to PyNWB and in from PyNWB to MatNWB

The round-trip tests are nearly identical to those in PyNWB. These tests write various NWBContainers to disk via nwbExport, read the containers back in via nwbRead, and then verify that the read containers exactly matches the written containers.

The out/in tests are similar to the round-trip tests but write containers to disk in MatNWB and verify they are read correctly by PyNWB and vice versa.

**Continuous integration and coverage reporting:**
The MatNWB test suite can be run by executing the nwbtest function in the root directory. In addition to running the test suite, the nwbtest function also writes a JUnit-style XML file containing the test results (testResults.xml) and a Cobertura-style XML file containing a code coverage report (coverage.xml).

I set up a multi-configuration Jenkins CI job to run nwbtest across multiple MATLAB releases (currently R2017a, R2017b, and R2018a prerelease). With the Cobertura Plugin, Jenkins also does a nice job of graphing and displaying the code coverage, but the coverage.xml should also be supported by codecov.io (currently used by PyNWB).


nclack:
Lawrence and I reviewed this yesterday and it looks fantastic.

PR #25, opened 2018-04-02
fix size check for matrix data

bendichter:


bendichter:
This changes the generated code in ElectricalSeries from:

```matlab
if ~isvector(val) || ~ismatrix(val)
    error('ElectricalSeries.data: val must have [1,2] dimensions');
end
```

to:

```matlab
if ~isvector(val) && ~ismatrix(val)
    error('ElectricalSeries.data: val must have [1,2] dimensions');
end
```

PR #28, opened 2018-04-28
Add script to update schema files and update schema files

jcfr:


bendichter:
I'd like to update the schema in matnwb and keep it up to date with pynwb and nwb-schema. Is that what this PR is trying to do? I tried generating the core using the schema in the latest pynwb dev branch:

```matlab
generateCore('path_to/pynwb/src/pynwb/data/nwb.namespace.yaml');
nwbRead('path_to/resting_state.nwb')
```

and it worked without an error :-)

bendichter:
nudge @nclack

PR #46, opened 2018-07-19
changes to DataStub.load

### LINKED ISSUE #44: reading only section of data
It looks like the way to read data is to call `DataStub.load` e.g. `file.acquisition.get('lfp').data.load`, which reads the entire matrix. Is it possible to only read a section of the data? e.g. [h5read](https://www.mathworks.com/help/matlab/ref/h5read.html)

bendichter:
* Order of args changed
* Uses MATALB built-in
* comments added

fix #44

### DIFF
--- a/+types/+untyped/DataStub.m
+++ b/+types/+untyped/DataStub.m
@@ -54,51 +54,19 @@ classdef DataStub
         
         %can be called without arg, with H5ML.id, or (dims, offset, stride)
         function data = load(obj, varargin)
-            fid = H5F.open(obj.filename);
-            did = H5D.open(fid, obj.path);
-            %             tid = H5D.get_type(did);
-            %             info = h5info(obj.filename, obj.path);
-            
-            switch length(varargin)
-                case 0
-                    sid = 'H5S_ALL';
-                    fsid = 'H5S_ALL';
-                case 1
-                    sid = varargin{1};
-                    fsid = sid;
-                otherwise
-                    dims = varargin{1};
-                    offset = varargin{2};
-                    stride = varargin{3};
-                    fsid = H5D.get_space(did);
-                    H5S.select_hyperslab(fsid, 'H5S_SELECT_SET',...
-                        offset,...
-                        stride,...
-                        [],... %everything is essentially one block
-                        dims);
-                    
-                    if all(dims == 1)
-                        sid = H5S.create('H5S_SCALAR');
-                    elseif any(dims == 0)
-                        sid = H5S.create('H5S_NULL');
-                    else
-                        % set Inf to unlimited
-                        dims(isinf(dims)) = ...
-                            H5ML.get_constant_value('H5S_UNLIMITED');
-                        sid = H5S.create_simple(numel(dims), dims, dims);
-                    end
-            end
-            
-            data = H5D.read(did, 'H5ML_DEFAULT', sid, fsid, 'H5P_DEFAULT');
-            if isstruct(data)
-                data = io.parseCompound(did, data);
-            end
-            if ~isempty(varargin) && ~isa(varargin{1}, 'H5ML.id')
-                %don't close a provided space id.
-                H5S.close(sid);
-            end
-            H5D.close(did);
-            H5F.close(fid);
+            %LOAD  Read data from HDF5 dataset.
+            %   DATA = LOAD() retrieves all of the data.
+            %
+            %   DATA = LOAD(START,COUNT) reads a subset of data.  START is 
+            %   the one-based index of the first element to be read.
+            %   COUNT defines how many elements to read along each dimension.  If a
+            %   particular element of COUNT is Inf, data is read until the end of the
+            %   corresponding dimension.
+            %
+            %   DATA = LOAD(START,COUNT,STRIDE) reads a strided subset of 
+            %   data.  STRIDE is the inter-element spacing along each
+            %   data set extent and defaults to one along each extent.
+            data = h5read(obj.filename, obj.path, varargin{:});
         end
         
         function refs = export(obj, fid, fullpath, refs)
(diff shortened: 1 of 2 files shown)


+types/+untyped/DataStub.m
```diff
@@ -54,51 +54,19 @@
...
-            did = H5D.open(fid, obj.path);
-            %             tid = H5D.get_type(did);
-            %             info = h5info(obj.filename, obj.path);
-            
-            switch length(varargin)
-                case 0
-                    sid = 'H5S_ALL';
-                    fsid = 'H5S_ALL';
-                case 1
-                    sid = varargin{1};
```
ln-vidrio: This change will remove the option for using HDF5 spaces as arguments.
Thankfully, only types.untyped.RegionView uses this internally, but committing this change now will definitely break its refresh() method.
Shouldn't be too big of a fix but I'm pretty sure regionview also uses 0-based indexing.

+types/+untyped/DataStub.m
```diff
@@ -54,51 +54,19 @@
...
-                        % set Inf to unlimited
-                        dims(isinf(dims)) = ...
-                            H5ML.get_constant_value('H5S_UNLIMITED');
-                        sid = H5S.create_simple(numel(dims), dims, dims);
-                    end
-            end
-            
-            data = H5D.read(did, 'H5ML_DEFAULT', sid, fsid, 'H5P_DEFAULT');
-            if isstruct(data)
-                data = io.parseCompound(did, data);
```
ln-vidrio: h5read won't automatically convert compound types to tables, they're always structs, so this change should be retained.

+types/+untyped/DataStub.m
bendichter: What if we just keep that option in?

```
+            %LOAD  Read data from HDF5 dataset.
+            %   DATA = LOAD() retrieves all of the data.
+            %
+            %   DATA = LOAD(SPACE) read data specified by HDF5 SPACE
+            %
+            %   DATA = LOAD(START,COUNT) reads a subset of data.  START is 
+            %   the one-based index of the first element to be read.
+            %   COUNT defines how many elements to read along each dimension.  If a
+            %   particular element of COUNT is Inf, data is read until the end of the
+            %   corresponding dimension.
+            %
+            %   DATA = LOAD(START,COUNT,STRIDE) reads a strided subset of 
+            %   data.  STRIDE is the inter-element spacing along each
+            %   data set extent and defaults to one along each extent.
```

+types/+untyped/DataStub.m
bendichter: could you show me an example of this feature?

+types/+untyped/DataStub.m
ln-vidrio: That's find, though there'll be an inconsistency between datastub and regionview indexing.  I think that's still okay though, I can change that later.

bendichter: Yeah I hear ya. Though I think it is also worth considering consistency between this package and MATLAB's native functions, which are generally one-indexed.

+types/+untyped/DataStub.m
ln-vidrio: read this file in (https://www.dropbox.com/s/eq73o0pzjprwggm/basic_example.nwb?dl=0) and load  `nwb.epochs.epochs.data`.  This will be a datastub load which you can breakpoint into.

bendichter:
The motivations of the PR were

1) Be more consistent with MATLAB (1-indexing)
2) Add documentation so a user could discover how to do this without looking at the code itself
3) Fix the `stride` issue. It wasn't working before and now it is.

This code deletes a lot of lower level code and replaces it with a high-level function. Presumably the low-level code did something, but whatever additional function it had was untested; making the replacement didn't break any of the tests. If functionality was broken, then this change reveals holes in the tests that should be filled. If it's easier to address the 3 motivations with smaller changes to the previous form of the code, then that's fine with me.

+types/+untyped/DataStub.m
bendichter: I don't see the issue. Seems right to me:

```
nwb.epochs.epochs.data.load

ans = 

  struct with fields:

     start_time: [2×1 double]
      stop_time: [2×1 double]
           tags: {2×1 cell}
     timeseries: [12×2 uint8]
    description: {2×1 cell}
```

bendichter: oh you want it to turn into a table?

ln-vidrio: Preferrably.  The easiest fix is to check if data is a struct (which only happens if it's compound) and call `io.parseCompound` on it.  By default, that's how compound data types are represented in matnwb.

bendichter: And the timeseries data is not correct. OK I see the issue

ln-vidrio: Yeah, MATLAB cuts a few corners when reading compound datasets.  References aren't even resolved, they just give the raw pointer value.
Again, just need to copy+paste this snippet below the `h5read` call:
```
if isstruct(data)
    data = io.parseCompound(did, data);
end
```

bendichter: Are the main advantages of tables that they enforce the same length for each column and that they print pretty? Am I missing something? The syntax seems pretty similar to structs besides that

ln-vidrio: Pretty printing and the fact that indexing into it using region references is a one-liner.
The tradeoff is that instantiation performance is terrible.

bendichter: huh ok. It looks like MATLAB's response to pandas dataframes

bendichter:
OK here's my attempt at a compromise. `nwb.epochs.epochs.data.load` works as expected on my side

ln-vidrio:
Looks good.  I'm working on tables and reference tests in the meantime.

PR #53, opened 2018-07-30
add requirements file

bendichter:
The requirements file allows you to specify dependencies of a package. This is the form of the file expected by [mpm](https://github.com/mobeets/mpm), which automatically finds downloads dependencies. My new loadTimeSeriesData function depends on [fastsearch](https://www.mathworks.com/matlabcentral/fileexchange/16976-fastsearch?s_tid=FX_rc1_behav).

### DIFF
--- a/+util/loadTimeSeriesData.m
+++ b/+util/loadTimeSeriesData.m
@@ -1,0 +1,75 @@
+function data = loadTimeSeriesData(timeseries, interval, downsample_factor, electrode)
+%LOADTIMESERIESDATA loads data within a time interval from a timeseries
+%
+%   DATA = loadTimeSeriesData(TIMESERIES, INTERVAL, DOWNSAMPLE_FACTOR)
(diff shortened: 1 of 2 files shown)


nclack:
@bendichter I'd like it better if we could just add fastsearch (and it's license) directly to the repo.  Maybe in a private folder adjacent to loadTimeSeriesData.  I'd rather avoid adding mpm as a dependency.

bendichter:
Yeah that's fine. It's still under active development and not quite ready
yet anyway

jcfr:
If you are venrorizing dependency (meaning copying their source code into this repo),  I suggest you adapt a systematic approach.

A script named `UpdateFastsearchFromUpstream.sh` could be added.  This is similar to the approach we are adding to sync the schema in pynwb/nwb-schema. See https://github.com/NeurodataWithoutBorders/pynwb/pull/506/files

PR #62, opened 2018-09-17
Add electrophysiology tutorial

bendichter:
Here is a working tutorial for ephys.

The html is generated from "Publish"ing `ecephys.m`. I'm not sure if we want to track the html or somehow generate it dynamically.

Also, I accidentally added then deleted an extraneous data file.

You can view the tutorial [here](http://htmlpreview.github.io/?https://github.com/NeurodataWithoutBorders/matnwb/blob/tutorials/tutorials/html/ecephys.html)


ln-vidrio:
Were these pages made using MATLAB publish?  I just want to know so that future documentation is consistent with this one.

PR #66, opened 2018-09-24
add util.blackrock and util.mworks conversion helpers

bendichter:
I made some helpers for converting from different common data acquisition system formats. How do you feel about putting them here?


ln-vidrio:
Could you move the files to a separate +contrib package instead of +util?  IMO third-party support should be separate from helper functions.

bendichter:
Ok sure

bendichter:
@ln-vidrio I have moved the files over to +contrib. Let me know if you would like any other changes.

Issue #74, opened 2018-10-10
support isodatetime dtype


oruebel:
@nclack @ln-vidrio this is a heads-up. When you sync up with the spec next time. The following pending PR https://github.com/NeurodataWithoutBorders/pynwb/pull/641  adds "isodatetime" as a dtype to the schema language. In HDF5 this simply maps to H5_TEXT (same as UTF-8 string). The main difference is that on the API side this implies that this should be a ISO 8601 date string. From a basic read/write perspective I think not much changes for MatNWB (since these are just string datasets), however, eventually this means you probably want a way to:
* represent these strings as Matlab datetime objects
* use Matlab datetime objects as inputs for isodatetime dtypes
* possibly additional validation on read/write

bendichter:
Under the new schema (currently merged in `pynwb/dev` but unmerged in `matnwb/master`) there is an error constructing an `NWBFile` object in matnwb and an error reading an nwbfile constructed using pynwb. Both errors have to do with the new format of `file_created_date` (and likely this will present in `session_start_time` as well).

### construction error
```matlab
nwb = nwbfile( 'source', 'acquired on rig2', ...
    'session_description', 'a test NWB File', ...
    'identifier', 'mouse004_day4', ...
    'session_start_time', datestr(date, 'yyyy-mm-ddTHH:MM:SS'), ...
    'file_create_date', datestr(now, 'yyyy-mm-ddTHH:MM:SS'));
```

gives the following error:
```matlab
Error using types.util.checkDtype (line 106)
Property `file_create_date` must be a .

Error in types.core.NWBFile/validate_file_create_date (line 302)
        val = types.util.checkDtype('file_create_date', '', val);

Error in types.core.NWBFile/set.file_create_date (line 188)
        obj.file_create_date = obj.validate_file_create_date(val);

Error in types.core.NWBFile (line 138)
        obj.file_create_date = p.Results.file_create_date;

Error in nwbfile (line 15)
            obj = obj@types.core.NWBFile(varargin{:});
```
### read error
file generated in python with:
```python
from pynwb import NWBHDF5IO, NWBFile
from datetime import datetime

nwbfile = NWBFile('source', 'session_description', 'identifier', datetime.now().astimezone())

with NWBHDF5IO('test_file.nwb', 'w') as io:
    io.write(nwbfile)
```

attempt to read in matlab gets the following error:
```matlab
>> generateCore('schema/core/nwb.namespace.yaml')
[...]

PR #75, opened 2018-10-16
[WIP] update schema

bendichter:
don't merge yet! Updates to schema need some attention: #74


bendichter:
substantial changes to ROIs as well that should be tested

Issue #43, opened 2018-07-17
Hardcoded Class Modifications in Schema Are not Checked for


ln-vidrio:
Main example being the `trials` Group under NWBFile.
```
- doc: 'Data about experimental trials'
    name: trials
    neurodata_type_inc: DynamicTable
    quantity: '?'
    datasets:
    - doc: The start time of each trial
      attributes:
        - name: description
          value: the start time of each trial
          dtype: text
          doc: Value is 'the start time of each trial'
      name: start
      neurodata_type_inc: TableColumn
      dtype: float
    - doc: The end time of each trial
      attributes:
        - name: description
          value: the end time of each trial
          dtype: text
          doc: Value is 'the end time of each trial'
      name: end
      neurodata_type_inc: TableColumn
      dtype: float
```
Expected behavior should be that NWBFile checks for the class along with any additions to the class itself.
At the current moment, only the class validation occurs.
This isn't necessarily breaking but validation is weaker, and this allows users to break from schema because matnwb never properly checks for these context-dependent values on export.

bendichter:
@ln-vidrio thanks for documenting this. Do you have any ideas for what would be the best approach to address this?

ln-vidrio:
@bendichter 
Open to other suggestions but here are my thoughts:
1) All generated classes inherit dynamicprops: https://www.mathworks.com/help/matlab/ref/dynamicprops-class.html
Probably the simplest method but would require establishing rules at runtime to determine if a given MATLAB property is an attribute, dataset, group or something else.

2) Generated classes "accumulate" all possible hardcoded changes and add them to the class implementation as optional parameters.
This one seems elegant but it imposes a global schema restriction, and there would need to be work done on going over every `neurodata_type_inc` and checking for available property additions.

yarikoptic:
a pressure point comment: is there a progress to get this old issue which exhibits itself widely to be addressed?

lawrence-mbf:
There was one more solution proposed aside from the ones above:

- Generate the equivalent of anonymous classes which include these changes (like <descriptor>_TimeSeries or something). This will most likely work for writes but reading will require something more advanced like some kind of pattern matching either in the anonymous class name or elsewhere. Not to mention how ugly the class names will end up being.

Open to suggestions though!

yarikoptic:
[...]

PR #78, opened 2018-11-05
Schema update

bendichter:
under the new schema, the `ecephys` tutorial breaks with the following error:

```matlab
Error using types.util.checkDtype (line 106)
Property `file_create_date` must be a .

Error in types.core.NWBFile/validate_file_create_date
(line 292)
        val = types.util.checkDtype('file_create_date',
        '', val);

Error in types.core.NWBFile/set.file_create_date (line
181)
        obj.file_create_date =
        obj.validate_file_create_date(val);

Error in types.core.NWBFile (line 135)
        obj.file_create_date = p.Results.file_create_date;

Error in nwbfile (line 15)
            obj = obj@types.core.NWBFile(varargin{:});

Error in ecephys (line 13)
nwb = nwbfile('session_description', 'a test NWB File',
...

### CODE REVIEW

**Comment by bendichter** [comment]
@ln-vidrio 
Are the optional columns for `DynamicTables` working as intended? For `trials` I get:

```matlab
trials = types.core.TimeIntervals( ...
    'colnames', {'correct'},...
    'description', 'trial data and properties', ...
    'id', types.core.ElementIdentifiers('data', 1:3));

>> trials

trials = 

  TimeIntervals with properties:

          start_time: []
           stop_time: []
                tags: []
          tags_index: []
          timeseries: []
    timeseries_index: []
            colnames: {'correct'}
         description: 'trial data and properties'
                  id: [1×1 types.core.ElementIdentifiers]
         tablecolumn: [1×1 types.untyped.Set]
          vectordata: [1×1 types.untyped.Set]
         vectorindex: [1×1 types.untyped.Set]
                help: 'A general epoch object'
```

the optional columns can be added at the root level:
```matlab
trials.start_time = types.core.TableColumn('data', [.1, 1.5, 2.5]);
```
but dynamic columns need to be added to `trials.tablecolumn`:
```matlab
>> trials.tablecolumn.set('correct', ...
    types.core.TableColumn('data', [0,1,0]));
```

If this is the most straightforward way to implement this that's fine. I just want to make sure it was intentional to have different syntaxes for adding optional and dynamic columns to a `DynamicTable`.

ln-vidrio:
I don't like it but I find them preferable to having a `tablecolumn` property.  There are currently no structures in place to actually validate that these properties exist so it might be better to keep them that way.
Eventually, I would actually want all dynamic properties to exhibit this behavior and not have a `tablecolumn` be visible to the user at all but that's unfortunately not feasible right now.

There is a bug if you instantiate using `start_time` as an input argument where it duplicates the value into `obj.start_time` and `obj.tablecolumn` as `start_time`.  I'll look into that.

ln-vidrio:
So with the new commit you should be able to instantiate TimeIntervals like this:
```
types.core.TimeIntervals(...
  'colnames', {'start_time', 'correct'},...
  'description', 'description',...
  'id', types.core.ElementIdentifiers('data', 1:3),...
  'start_time', types.core.TableColumn('data', [.1, 1.5, 2.5]),...
  'correct', types.core.TableColumn('data', [0, 1, 0]))
```
and the properties should be in the right place.
```
ans = 

  TimeIntervals with properties:

          start_time: [1×1 types.core.TableColumn]
           stop_time: []
                tags: []
          tags_index: []
          timeseries: []
    timeseries_index: []
            colnames: {'start_time'  'correct'}
         description: 'description'
                  id: [1×1 types.core.ElementIdentifiers]
         tablecolumn: [1×1 types.untyped.Set]
          vectordata: [1×1 types.untyped.Set]
         vectorindex: [1×1 types.untyped.Set]
                help: 'A general epoch object'
```
where `tablecolumn` looks like this:
```
ans = 

  Set with properties:

    correct: [types.core.TableColumn]
```

bendichter:
got it. Cool. I'm also dealing with another issue right now, but I have to admit I don't fully understand the python changes yet, so that may be the source of the issue. I'm trying to add spike times to the units table. I can add the `VectorData` like this:
```matlab
units.spike_times = types.core.VectorData('data', [1,2,3,4,5]);
```
but then I don't see how I would indicate what spikes are associated with what rows of the table.

ln-vidrio:
My assumption without looking at anything else would be that each element in the VectorData corresponds to a row so each value in VectorData is a spike time.  The `doc` for `spike_times` says it indicates `the spike times for each unit` which I would assume is each "row" in the dynamic table.

bendichter:
It's definitely supposed to store more than one float per row. I'll investigate how this is done in pynwb

ln-vidrio:
Perhaps `spike_times_index` is supposed to correspond to each row and reference parts of `spike_times`?

bendichter:
aha! Yes that's it. It's using the matnwb underscore expansion syntax! Of course. Thanks for the help.

ln-vidrio:
It's actually not, but the `doc` property describes it as such.  Here is the yaml schema snippet for Units:
```
  datasets:
  - neurodata_type_inc: VectorIndex
    name: spike_times_index
    doc: the index into the spike_times dataset
    quantity: '?'
  - neurodata_type_inc: VectorData
    name: spike_times
    doc: the spike times for each unit
    quantity: '?'
```
So in raw HDF5, spike_times_index is a dataset under the `Units` group which is adjacent to the spike_times dataset.  I believe this is the replacement for Region References (types.untyped.RegionView) so it should be used in a similar way.

bendichter:
Is it true that order to get `nwb.trials` to pass validation on `nwbExport`, you need to add description to each column?

```matlab
date = datetime(2018, 3, 1, 12, 0, 0);
session_start_time = datetime(date,'Format','yyyy-MM-dd''T''HH:mm:SSZZ',...
    'TimeZone','local');
nwb = nwbfile( 'source', 'acquired on rig2', ...
    'session_description', 'a test NWB File', ...
    'identifier', 'mouse004_day4', ...
    'session_start_time', session_start_time);

trials = types.core.TimeIntervals( ...
    'colnames', {'correct','start_time','stop_time'},...
    'description', 'trial data and properties', ...
    'id', types.core.ElementIdentifiers('data', 1:3),...
    'start_time', types.core.TableColumn('data', [.1, 1.5, 2.5],...
        'description','hi'),...
    'stop_time', types.core.TableColumn('data', [1., 2., 3.],...
        'description','hi'),...
    'correct', types.core.TableColumn('data', [false,true,false],...
        'description','my description'));

nwb.intervals.set('trials', trials)

nwbExport(nwb, 'trials_test.nwb')
```

ln-vidrio:
Yes, `description` under TableColumn is required as it is a non-optional attribute:
```
- neurodata_type_def: TableColumn
  neurodata_type_inc: NWBData
  doc: The attributes specified here are included in all interfaces.
  attributes:
  - name: help
    dtype: text
    doc: Value is 'One of many columns that can be added to a DynamicTable'.
    value: One of many columns that can be added to a DynamicTable
  - name: description
    dtype: text
    doc: A short description of what this column stores
```

bendichter:
Yes, but it's not ideal that this forces the user to overwrite the description supplied by the schema:
https://github.com/NeurodataWithoutBorders/matnwb/blob/e034ed0db38ac594a4bc657031d9de89bbfa4c82/schema/core/nwb.epoch.yaml#L12-L29

ln-vidrio:
Right, but that definition runs afoul of #43.  I agree that's a pain to actually work with, though I'm unsure if that actually breaks compatibility with pynwb.  I'll look into it as I have an idea in mind but no guarantees for right now.

bendichter:
It don't think it will break compatibility with pynwb, I just wanted to see if there was a way of getting this data in without overwriting the description field.

ln-vidrio:
I think the most I can do is make the default value match as close as possible so long as it isn't overwritten by user data.  On second thought, that might not be terribly useful.

bendichter:
How would you feel about having each of those optional columns be instantiated with a`TableColumn` with empty data e.g. `nwb.trials.start_time = types.core.TableColumn('description','start time of the epoch, in seconds')` and then having the user add data with `trials.start_time.data = [1,2,3]`?

ln-vidrio:
Well that's what I'm considering, but the user would have to know that beforehand (further separating optional Tablecolumns and dynamic ones which still require `description`) and it would split the instantiation of the TableColumn to two calls (one for instantiation of the object and another to set the data).  It might not be that bit of a deal but it would look clunky.

That said, it would help as a form of documentation and would probably be more helpful than an empty array.

bendichter:
In my mind separating the syntax for optional and dynamic columns isn't so bad. In fact, I wouldn't be opposed to:
```matlab
nwb.trials.start_time.data = [1,2,3]
nwb.trials.add_column(name,description,data)
```
That would be closer to how pynwb does things.

ln-vidrio:
Ideally that would be the case, though again that requires some work around how they currently work.  I'll look into seeing if I can get the hardcoded data added in by default.

ln-vidrio:
looks like a naming collision in  /general/extracellular_ephys/electrodes. The DynamicTable attribute `description` is being confused with a column named `description`.
They're both considered properties when read in so nwbRead confused one for the other.  This only showed up now because we disable parsing constrained values to static properties in the previous commit.  Before then, the `description` column was properly deleted before the other static properties were parsed.

bendichter:
@ln-vidrio ok I have the tutorial back up and working with the new schema

ln-vidrio:
Is there anything else that needs to be done?

PR #80, opened 2018-11-10
[WIP] add enhanced subject

bendichter:


bendichter:
@ln-vidrio 

I'm playing with including more information in the Subject object, and I am running into an issue that I think may be a bug in matnwb. On this branch, when I run `types.core.Subject` I get:

```matlab
Error using types.util.checkDtype (line 107)
Property `surgeries` must be a types.core.Surgeries.

Error in types.core.Subject/validate_surgeries (line 134)
        val = types.util.checkDtype('surgeries', 'types.core.Surgeries',
        val);

Error in types.core.Subject/set.surgeries (line 99)
        obj.surgeries = obj.validate_surgeries(val);

Error in types.core.Subject (line 64)
        obj.surgeries = p.Results.surgeries;
```

`Surgeries` is optional (`quantity=: '?'`), so I don't think I should be getting this error.

ln-vidrio:
119c4e11935ab5fff5fd2ca75a58e1f1fcde22dd
This seems to fix it.  Requires generating again.

bendichter:
@ln-vidrio thanks! That works but now I think I found another quantity bug.

The following code:
```matlab
subject = types.core.Subject('surgeries',types.core.Surgeries());

file = nwbfile('session_description', 'a test NWB File', ...
'identifier', 'blockname', ...
'session_start_time', datestr(now, 'yyyy-mm-dd HH:MM:SS'), ...
'file_create_date', datestr(now, 'yyyy-mm-dd HH:MM:SS'),...
'general_subject',subject);

nwbExport(file, 'test.nwb')
```
runs but should throw an error.

`types.core.Surgery` is defined as a group of `types.core.Surgeries` with `quantity: +`, which means `Surgeries` must contain "one or more" `Surgery` objects. This code creates `Subject.surgeries.surgery` with an empty set, which is a violation of the schema. I would prefer if this was caught in the constructor, but it could also be caught during `nwbExport` (it currently is not).

ln-vidrio:
8a914802e655a850d3f40aa7c9b62c4d0bebddbd should be correct.  I opted against checking on construction since these are also called when reading nwbfiles.

bendichter:
Yeah that's reasonable

yarikoptic:
is it really still WiP @bendichter or might want to be closed (or converted to draft at least)?

PR #81, opened 2018-11-29
update schema

bendichter:
We're trying to stay away from schema-breaking changes, so there shouldn't be any problems with this one. The tests and tutorials seem to run without problem


ln-vidrio:
Looks good on my end.

PR #84, opened 2018-12-17
Rmv table column

bendichter:


bendichter:
tablecolumn as replaced with vectordata, since they ended up being identical. I have updated the utilities and tutorials accordingly.

ln-vidrio:
Sorry for not noticing this pull request sooner.  I didn't realize it made changes to internal code.
We should do something about that datetime validation, I can't imagine it being maintainable.  Perhaps use an external library like [https://www.mathworks.com/matlabcentral/fileexchange/39389-iso-8601-date-string-to-serial-date-number](https://www.mathworks.com/matlabcentral/fileexchange/39389-iso-8601-date-string-to-serial-date-number)?
Otherwise looks okay.

PR #87, opened 2018-12-31
add data stub region read test

bendichter:


ln-vidrio:
The rule I've been generally following is that system tests should be PyNWB class-specific tests while unit tests are functionality-specific tests.
IMO, this should go in unit tests as it doesn't play well with the RoundTripTest class.  You can take a look at the +tests/+unit test functions; they use the Function-Based Unit Tests testing framework (https://www.mathworks.com/help/matlab/function-based-unit-tests.html).

If you wish I can also convert this over.

ln-vidrio:
This is done but needs to be checked with #88 since that will change how load works.

PR #88, opened 2018-12-31
add easier region read method for data stubs

bendichter:


ln-vidrio:
Should probably default load() to this behavior and have the current load() become load_raw or something.  I doubt people prefer the raw read method.  That said, I believe you mentioned that it would break tutorials?

bendichter:
Yeah I agree. And yes it would break tutorials and some other tools I've made but I can fix those

ln-vidrio:
Okay, there'll be some minor internal fixes as well so I'll look into this.

PR #92, opened 2019-01-15
[WIP] changes to schema

bendichter:
currently `generateCore` throws error:

```matlab
Dot indexing is not supported for variables of this type.

Error in file.fillValidators>fillUnitValidation (line 91)
            sg_namespace = namespacereg.getNamespace(sg.type).name;

Error in file.fillValidators (line 12)
        validationBody = fillUnitValidation(nm, prop, namespacereg);

Error in file.fillClass (line 87)
validatorFcns = file.fillValidators(allprops, classprops, namespace);

Error in file.writeNamespace (line 15)
        fwrite(fid, file.fillClass(k, namespace, processed, ...

Error in generateExtension (line 42)
file.writeNamespace(extmap(extSchema.name));

Error in generateCore (line 22)
generateExtension(core);
 
Reference to non-existent field 'internal'.
```

### DIFF
--- a/schema/core/nwb.base.yaml
+++ b/schema/core/nwb.base.yaml
@@ -168,9 +168,27 @@ groups:
         such as Volts. If the actual data is stored in millivolts, the field ''conversion''
         below describes how to convert the data to the specified SI unit.'
     dims:
-    - num_times
+    - - num_times
+    - - num_times
+      - num_DIM2
+    - - num_times
+      - num_DIM2
+      - num_DIM3
+    - - num_times
+      - num_DIM2
+      - num_DIM3
+      - num_DIM4
     shape:
-    - null
+    - - null
+    - - null
+      - null
+    - - null
+      - null
+      - null
+    - - null
+      - null
+      - null
+      - null
   - name: starting_time
     dtype: float64
     doc: 'The timestamp of the first sample. COMMENT: When timestamps are uniformly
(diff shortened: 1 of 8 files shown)


bendichter:
This error appears to be in constructing Devices.

ln-vidrio:
The inclusion of the `LabMetaData` type has broken elision for general.  I'll see if I can gracefully split that into multiple properties.

bendichter:
I see. It's sort of a funny type because it's sole purpose is to be extended, but it saves people from having to extend NWBFile itself, which I imagine would cause bigger problems. Thanks for handling this. I'm also pushing some changes that I know will be necessary (though I can't test them while `generateCore` is broken)

bendichter:
@ln-vidrio any progress here?

ln-vidrio:
I'm fixing tests now but it's essentially a schema reading adjustment so I'd imagine it will also break most if not all tests and utilities.

bendichter:
yikes. OK... thanks for getting this in

ln-vidrio:
So I'll summarize script changes:
Due to changes in elision rules, the properties for any class with Sets have been restructured.
Basically, if a property is "missing" it may've been "hoisted" as its parent property.  So you can't find "general_labmetadata" as a property because it's been hoisted into the "general" property.  Hopefully, this will only cause syntax errors in the scripts.

Issue #106, opened 2019-02-05
TimeSeriesIOTest not passing


bendichter:
```matlab
================================================================================
Verification failed in tests.system.TimeSeriesIOTest/testOutToPyNWB.

    ----------------
    Test Diagnostic:
    ----------------
    
    ======================================================================
    FAIL: testInFromMatNWB (PyNWBIOTest.TimeSeriesIOTest) (container_type='TimeSeries', nwbfield='data')
    ----------------------------------------------------------------------
    Traceback (most recent call last):
      File "/Users/bendichter/dev/matnwb/+tests/+system/PyNWBIOTest.py", line 99, in assertContainerEqual
        self.assertEqual(f1, f2)
    AssertionError: <HDF5 dataset "data": shape (10,), type "<f8"> != [100, 110, 120, 130, 140, 150, 160, 170, 180, 190]
    
    ======================================================================
    FAIL: testInFromMatNWB (PyNWBIOTest.TimeSeriesIOTest) (container_type='TimeSeries', nwbfield='timestamps')
    ----------------------------------------------------------------------
    Traceback (most recent call last):
      File "/Users/bendichter/dev/matnwb/+tests/+system/PyNWBIOTest.py", line 99, in assertContainerEqual
        self.assertEqual(f1, f2)
    AssertionError: <HDF5 dataset "timestamps": shape (10,), type "<f8"> != array([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
    
    ----------------------------------------------------------------------
    Ran 1 test in 0.038s
    
    FAILED (failures=2)

    ------------------
    Stack Information:
    ------------------
    In /Users/bendichter/dev/matnwb/+tests/+system/PyNWBIOTest.m (PyNWBIOTest.testOutToPyNWB) at 15
================================================================================
.
================================================================================
Verification failed in tests.system.TimeSeriesIOTest/testInFromPyNWB.

    ----------------
    Test Diagnostic:
    ----------------
    Values for property 'timestamps_interval' are not equal

    ---------------------
    Framework Diagnostic:
    ---------------------
    verifyEqual failed.
    --> Classes do not match.
        
        Actual Class:
            int64
        Expected Class:
            int32
    
    Actual Value:
      int64
    
       1
    Expected Value:
      int32
    
       1

    ------------------
    Stack Information:
    ------------------
[...]

Issue #105, opened 2019-02-05
PhotonSeriesIOTest not passing


bendichter:
```matlab
================================================================================
Verification failed in tests.system.PhotonSeriesIOTest/testOutToPyNWB.

    ----------------
    Test Diagnostic:
    ----------------
    /Users/bendichter/dev/pynwb/src/pynwb/image.py:80: DeprecationWarning: bits_per_pixel is no longer used
      warnings.warn("bits_per_pixel is no longer used", DeprecationWarning)
    E
    ======================================================================
    ERROR: testInFromMatNWB (PyNWBIOTest.PhotonSeriesIOTest)
    ----------------------------------------------------------------------
    Traceback (most recent call last):
      File "/Users/bendichter/dev/matnwb/+tests/+system/PyNWBIOTest.py", line 33, in testInFromMatNWB
        matcontainer = self.getContainer(matfile)
      File "/Users/bendichter/dev/matnwb/+tests/+system/PyNWBIOTest.py", line 170, in getContainer
        return file.get_acquisition(self.container.name)
    AttributeError: 'NoneType' object has no attribute 'name'
    
    ----------------------------------------------------------------------
    Ran 1 test in 0.061s
    
    FAILED (errors=1)

    ------------------
    Stack Information:
    ------------------
    In /Users/bendichter/dev/matnwb/+tests/+system/PyNWBIOTest.m (PyNWBIOTest.testOutToPyNWB) at 15
================================================================================
.
================================================================================
Error occurred in tests.system.PhotonSeriesIOTest/testInFromPyNWB and it did not run to completion.

    ---------
    Error ID:
    ---------
    'MATLAB:getReshapeDims:notSameNumel'

    --------------
    Error Details:
    --------------
    Error using reshape
    To RESHAPE the number of elements must not change.
    
    Error in tests.system.RoundTripTest/verifyContainerEqual (line 73)
                            trueval = reshape(trueval, size(val2));
    
    Error in tests.system.PyNWBIOTest/testInFromPyNWB (line 28)
          testCase.verifyContainerEqual(pycontainer, matcontainer);
================================================================================
```

bendichter:
depends on https://github.com/NeurodataWithoutBorders/pynwb/issues/815

PR #107, opened 2019-02-05
fix IO tests

bendichter:
partial fix to #105, #106


ln-vidrio:
What does partial fix mean in this case?

bendichter:
There were several errors within each of these tests, and I was able to fix a few of them quickly, but there were others that I wasn't sure the best way to fix. I wasn't able to fix any of the tests all the way. If I write "fix ##" and this gets merged, it will automatically close an issue, and I didn't want that to happen, so I just wanted to indicate that these changes do not entirely fix the issues.

Issue #118, opened 2019-03-08
Header loading memory issue


mpompolas:
I try to load the header of a 10GB NWB file on my old laptop (4 GB RAM) and I run into a memory issue as shown below:

![screen shot 2019-03-07 at 9 20 26 pm](https://user-images.githubusercontent.com/23224563/54003578-a553c400-4120-11e9-9836-d9629d74ed95.png)

Is this command supposed to load data as well?

Thanks,
Konstantinos

bendichter:
Can you please copy/paste the actual error message and a link to the file?

mpompolas:
Link:

https://drive.google.com/file/d/1ceoaieEpn7NVmU36o-yRTWwJFYjjWdJQ/view



------------------------------------------------------------------------------------------------------
Error using hdf5lib2
Requested 64x50461375 (6.0GB) array exceeds maximum array size preference. Creation
of arrays greater than this limit may take a long time and cause MATLAB to become
unresponsive. See array size limit or preference panel for more information.

Error in H5D.read (line 51)
buf = H5ML.hdf5lib2('H5Dread',varargin{:});

Error in types.untyped.DataStub/load_h5_style (line 58)
                data = H5D.read(did, 'H5ML_DEFAULT', varargin{1}, ...

Error in types.untyped.DataStub/load (line 109)
                data = obj.load_h5_style(sid);

Error in types.util.checkDtype (line 70)
            val = val.load(1);

Error in types.core.ElectricalSeries/validate_data (line 37)
        val = types.util.checkDtype('data', 'numeric', val);

Error in types.core.TimeSeries/set.data (line 96)
        obj.data = obj.validate_data(val);

Error in types.core.TimeSeries (line 67)
        obj.data = p.Results.data;

Error in types.core.ElectricalSeries (line 16)
        obj = obj@types.core.TimeSeries(varargin{:});

Error in io.parseGroup (line 67)
    parsed = eval([typename '(kwargs{:})']);

Error in io.parseGroup (line 28)
    subg = io.parseGroup(filename, g_info);

Error in io.parseGroup (line 28)
    subg = io.parseGroup(filename, g_info);

Error in io.parseGroup (line 28)
    subg = io.parseGroup(filename, g_info);

Error in io.parseGroup (line 28)
    subg = io.parseGroup(filename, g_info);

Error in nwbRead (line 20)
    nwb = io.parseGroup(filename, info);

ln-vidrio:
[...]

PR #119, opened 2019-03-08
DataStub unnecessary allocation on indexing

ln-vidrio:
potential fix for #118


bendichter:
@ln-vidrio I generally differ to you on these types of things. If you think this is the best solution and the tutorials still run, I'll support it.

PR #127, opened 2019-04-16
add ophys tutorial

bendichter:


bendichter:
@ln-vidrio, I'm running into some problems with an error on `nwbExport`. I may have made a mistake but I haven't been able to find it. Can you look at this for me?

ln-vidrio:
What's the error message?

ln-vidrio:
The issue is that io.resolvePath isn't checking nested Set objects.  I'll take a look into that.

PR #134, opened 2019-05-06
Renamed Example Links and added Summaries

ln-vidrio:


bendichter:
@ln-vidrio perfect, thanks!

PR #136, opened 2019-05-10
Flip Datastub Dims

ln-vidrio:
For better load() API consistency


bendichter:
this seems like it'll probably break some stuff. I'm nervous about merging this before the hackathon

ln-vidrio:
Yeah, the only reason I wanted to do this was because of how weird it looked when I was testing it for my presentation.  I'm okay with merging after hackathon as well.

ln-vidrio:
Now that the hackathon's over, are you alright with this change?

bendichter:
Have you checked to make sure the demos still run with these changes?

ln-vidrio:
Hm, it looks like some changes after the hackathon made the tests fail too.  I'll let you know when these are all working.

ln-vidrio:
Tests pass.
The conversion and basicUsage trials look fine.  I ran the icephys and ecephys tutorials and they seemed alright.  I can run ophys but only after removing the Statistics Toolbox dependency so I'm unsure as to the validity of the data.

PR #147, opened 2019-05-16
add optional column param

bendichter:


ln-vidrio:
This is a change to your script isn't it?  Should it be merged now?

PR #168, opened 2019-09-17
Add details to the setup instructions

oruebel:
Add details on how to retrieve the MatNWB sources and NWB schema to the setup instructions


bendichter:
thanks @oruebel I think this is an improvement, and particularly helpful for MATLAB users who are generally less likely to be familiar with git.

PR #174, opened 2019-12-27
update schema paths, and update README and generateCore to reflect this

### LINKED ISSUE #173: update paths
`generateCore` currently expects paths `schema/core/nwb.namespace.yaml` and `schema/common/namespace.yaml`. 

`git clone https://github.com/NeurodataWithoutBorders/nwb-schema.git`

creates filepath `nwb-schema/core/nwb.namespace.yaml` and the core is missing because you need to call `git clone` in recursive mode. If you correct that, calling instead 

`git clone --recursive https://github.com/NeurodataWithoutBorders/nwb-schema.git`

you also get path `nwb-schema/hdmf-common-schema/common/namespace.yaml`. In order to update our code to the current schema, we need to 

1. Change the git clone call to be recursive
2. Update the filepaths in `generateCore`

bendichter:
fix #173


ln-vidrio:
Is it alright if I roll in some minor README changes here as well?  These are primarily wording changes to imply that the nwb-schema is the primary source of the updated schema version now.

I will also add the MATLAB File Exchange badge to the download section.

bendichter:
@ln-vidrio sounds good to me!

PR #175, opened 2019-12-27
nwb-schema 2.2.0

bendichter:
This branch keeps up with changes in the dev branch of nwb-schema. As of right now, everything works, but since the latest changes to nwb-schema are not yet released, we should not fold this into the master branch yet. Once nwb-schema 2.2.0 is released, we can update this, merge it into master, and release it.


ln-vidrio:
Moved to #186 to support 2.2.1

PR #179, opened 2019-12-31
Enhancement/iterative write

bendichter:


ln-vidrio:
@bendichter When you get time, please pull to tip of branch and run ecephys tests again.

bendichter:
@ln-vidrio looks like it works

PR #197, opened 2020-04-01
Simplify DataPipe Construction

ln-vidrio:
Fixes #195

PR #209, opened 2020-05-05
DataPipe Tutorials Added

IvanSmal:
Created the tutorials for compressing data using dataPipe


ln-vidrio:
Is this intended to be merged after #197 or does it also work on master?  Just want to make sure I'm making the right assumptions with merging here.

bendichter:
This has been merged via a direct commit. Thanks @IvanSmal!

PR #215, opened 2020-05-13
Add Nwb search method for specific types.

ln-vidrio:
`nwb.searchFor(typename)` searches the nwb object recursively for any objects that satisfy any one of the following criteria:

1. object is a class/subclass of class specified by `typename` (case sensitive)
2. object's class/subclass name ends with `typename` (case insensitive)

It then returns a `containers.Map` object that consists of a `nwb.resolve()`-able class mapped to the found object.
**Note** the provided path may not be accurate to its actual HDF5 path. Its only guarantee is that the path is `resolve()`-able.

### DIFF
--- a/+io/resolvePath.m
+++ b/+io/resolvePath.m
@@ -12,6 +12,8 @@ errmsg = 'Could not resolve path `%s`.';
 while ~isempty(tokens)
     if isa(o, 'types.untyped.Set')
         [o, tokens] = resolveSet(o, tokens);
+    elseif isa(o, 'types.untyped.Anon')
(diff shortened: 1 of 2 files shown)


jens-k:
I tried this on several different datasets and it always worked. Thanks!!
As discussed elsewhere, a 'strict' flag would be nice, to keep it from finding subclasses.

ln-vidrio:
@jens-k The search by default should only search for names now.
If you still wish to expand your search to include subclasses, then add the optional flag "includeSubClasses" (case insensitive).

jens-k:
Seems to work well!

PR #218, opened 2020-05-14
BoundPipe now subclasses DataStub

ln-vidrio:
Can now load data from BoundPipes.


bendichter:
What is a BoundPipe?

ln-vidrio:
@bendichter This is an internal type in DataPipe that represents an association with data written to file. Internally, DataPipe uses this type to indicate to itself that it has been written to file.
Since this type indicates data on disk, it should make sense that one should be able to read from it like a DataStub. In this case, BoundPipe can now load its own data like with DataStubs.

lvsltz:
Just to let you know I tested it with my script and everything looks ok :). Thanks

PR #223, opened 2020-05-15
Class and Namespace files now saved in tempdir

### LINKED ISSUE #220: Create +types / namespaces folders in Matlab's temporary folder
When calling nwbRead, there are two folders created in the pwd. This can be unexpected to the user, especially when the function is called somewhere under the hood (e.g., when reading in an NWB file using a third-party software).

As a way to avoid creating folders in somewhat arbitrary locations on the user's computer, they could be created by default in matlabs designated `tempdir` .

See a short discussion on this here: 
https://nwb-users.slack.com/archives/C5XKC14L9/p1589498856181700

lawrence-mbf:
Fixes #220 
Now returns the tempdir instead of the user's working directory.

### DIFF
--- a/+file/writeNamespace.m
+++ b/+file/writeNamespace.m
@@ -2,7 +2,7 @@ function writeNamespace(namespaceName)
 %check/load dependency namespaces
 Namespace = schemes.loadNamespace(namespaceName);
 
-path = fullfile(misc.getWorkspace(), '+types', ['+' Namespace.name]);
+path = fullfile(misc.getClassDir(), '+types', ['+' Namespace.name]);
 if exist(path, 'dir') == 7
     rmdir(path, 's');
 end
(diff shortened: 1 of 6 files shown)


jens-k:
Hey, this looks like a great fix. Is it gonna be merged soon?

ln-vidrio:
@jens-k Unfortunately I would not put too much stock into this being merged anytime soon. There is still the issue of creating a permanent path in MATLAB which may require implementing a programmatic file editor. Furthermore, we are currently on maintenance mode due to our current contract for this half year. If you wish, you can fork off of this branch and we can review a workable PR in the future but that is the most I can do right now, unfortunately.

lawrence-mbf:
Closed due to staleness. The current default usage is to install the types files in the install directory with an option to save in some `savedir` option. Perhaps there should be a special option like `savedirtemp` to force saving there?

PR #226, opened 2020-05-28
add doc to DataPipe class

### LINKED ISSUE #50: data compression
Is there any way to use matnwb to tell HDF5 to apply compression to a dataset?

Breaking it down, the following will be needed for implementation:

 - [x] decide on user interface. My vote would be to mirror pynwb and create a data object that can hold information about compression parameters. Would `DataStub` work for this?
 - [x] implement data compression when writing the data to the HDF5 file according to the parameters provided. See https://github.com/NeurodataWithoutBorders/matnwb/issues/50#issuecomment-414756686
 - [x] roundtrip tests.
 - [ ] documentation
    - [ ] informative function comments
    - [x] including in a tutorial

bendichter:
close #50

### DIFF
--- a/+types/+untyped/DataPipe.m
+++ b/+types/+untyped/DataPipe.m
@@ -1,6 +1,36 @@
 classdef DataPipe < handle
-    %DATAPIPE Summary of this class goes here
-    %   Detailed explanation goes here
+    %DATAPIPE gives advanced write directions to HDF5 for a dataset for
+    %chunking, compression, and iterative write.
+    %   DATAPIPE directs HDF5 to use chunking and GZIP compression when 
+    %   saving the dataset. The chunk size is automatically determined and
+    %   the compression level is 3 by default.
+    %
+    %	DATAPIPE(..., 'data', DATA) Write DATA in the Dataset.
+    %
+    %   DATAPIPE(..., 'maxSize', MAXSIZE) Sets the maximum size of the HDF5
+    %   Dataset. Unless using iterative writing, this should match the size
+    %   of Data. To append data later, use the maxSize for the full 
+    %   dataset. You can use Inf for a value of a dimension if you do not
+    %   know its final size.
+    %
+    %   DATAPIPE(..., 'axis', AXIS) Set which axis to increment when
+    %   appending more data.
+    %
+    %   DATAPIPE(..., 'dataType', DATATYPE) Sets the type of the data. This
+    %   must be a numeric data type. Useful to include when using iterative
+    %   write to append data as the appended data must be the same data
+    %   type. If data is provided and dataType is not, the dataType is
+    %   inferred from the provided data.
+    %
+    %   DATAPIPE(..., 'chunkSize', CHUNKSIZE) Sets chunk size. Must be less
+    %   than maxSize. If not provided, the CHUNKSIZE will be automatically
+    %   determined.
+    %
+    %   DATAPIPE(..., 'compressionLevel', COMPRESSIONLEVEL) sets a
+    %   GZIP compression level over than the default of 3
+    %
+    %   DATAPIPE(..., 'offset', OFFSET) Axis offset of dataset to append.
+    %   May be used to overwrite data.
     
     properties (SetAccess = private)
         internal;
--- a/+types/+untyped/DataPipe.m
+++ b/+types/+untyped/DataPipe.m
@@ -1,36 +1,47 @@
 classdef DataPipe < handle
-    %DATAPIPE gives advanced write directions to HDF5 for
-    %chunking, compressing, and iteratively writing a dataset.
+    %DATAPIPE gives advanced write directions to HDF5 for a dataset for
+    %chunking, compression, and iterative write.
     %   DATAPIPE directs HDF5 to use chunking and GZIP compression when 
     %   saving the dataset. The chunk size is automatically determined and
     %   the compression level is 3 by default.
     %
-    %	DATAPIPE(..., 'data', DATA) Write DATA in the Dataset.
+    %	DATAPIPE(..., 'data', DATA) Preload DATA in the Dataset. This can
+    %	be omitted if the DATA will be appended later
     %
     %   DATAPIPE(..., 'maxSize', MAXSIZE) Sets the maximum size of the HDF5
-    %   Dataset. Unless using iterative writing, this should match the size
-    %   of Data. To append data later, use the maxSize for the full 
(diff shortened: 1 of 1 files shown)


+types/+untyped/DataPipe.m
```diff
@@ -1,6 +1,36 @@
...
-    %   Detailed explanation goes here
+    %DATAPIPE gives advanced write directions to HDF5 for a dataset for
+    %chunking, compression, and iterative write.
+    %   DATAPIPE directs HDF5 to use chunking and GZIP compression when 
+    %   saving the dataset. The chunk size is automatically determined and
+    %   the compression level is 3 by default.
+    %
+    %	DATAPIPE(..., 'data', DATA) Write DATA in the Dataset.
+    %
+    %   DATAPIPE(..., 'maxSize', MAXSIZE) Sets the maximum size of the HDF5
```
ln-vidrio, line 10: I suggest adding that `maxSize` can be inferred from a provided `data` parameter if it is provided, instead of suggesting that `maxSize` "should match" as if it were a user requirement (which it is but only if you don't provide `data`.

ln-vidrio: Furthermore, the default value should at least be documented. If not provided and there is no `data`, then the assumed maximum shape is an infinite vector.

+types/+untyped/DataPipe.m
```diff
@@ -1,6 +1,36 @@
...
+    %   Dataset. Unless using iterative writing, this should match the size
+    %   of Data. To append data later, use the maxSize for the full 
+    %   dataset. You can use Inf for a value of a dimension if you do not
+    %   know its final size.
+    %
+    %   DATAPIPE(..., 'axis', AXIS) Set which axis to increment when
```
ln-vidrio, line 16: Same as above with MAXSIZE, the default if not provided should be noted as the first dimension.

+types/+untyped/DataPipe.m
```diff
@@ -1,6 +1,36 @@
...
+    %   write to append data as the appended data must be the same data
+    %   type. If data is provided and dataType is not, the dataType is
+    %   inferred from the provided data.
+    %
+    %   DATAPIPE(..., 'chunkSize', CHUNKSIZE) Sets chunk size. Must be less
+    %   than maxSize. If not provided, the CHUNKSIZE will be automatically
+    %   determined.
+    %
+    %   DATAPIPE(..., 'compressionLevel', COMPRESSIONLEVEL) sets a
+    %   GZIP compression level over than the default of 3
```
ln-vidrio, line 30:
```suggestion
    %   DATAPIPE(..., 'compressionLevel', COMPRESSIONLEVEL) sets the
    %   GZIP compression level (see the <a href="matlab:web('http://portal.hdfgroup.org/display/HDF5/H5P_SET_DEFLATE','-browser')">H5P_SET_DEFLATE documentation</a> for more information on compression levels).
    %   The default compression level is 3.
```

+types/+untyped/DataPipe.m
```diff
@@ -1,6 +1,36 @@
 classdef DataPipe < handle
-    %DATAPIPE Summary of this class goes here
-    %   Detailed explanation goes here
+    %DATAPIPE gives advanced write directions to HDF5 for a dataset for
+    %chunking, compression, and iterative write.
```
ln-vidrio, line 3:
```suggestion
    %DATAPIPE gives advanced write directions to HDF5 for
    %chunking, compressing, and iteratively writing a dataset.
```

+types/+untyped/DataPipe.m
ln-vidrio: Should `hasShuffle` be included here?

+types/+untyped/DataPipe.m
```diff
@@ -1,6 +1,36 @@
...
+    %   DATAPIPE directs HDF5 to use chunking and GZIP compression when 
+    %   saving the dataset. The chunk size is automatically determined and
+    %   the compression level is 3 by default.
+    %
+    %	DATAPIPE(..., 'data', DATA) Write DATA in the Dataset.
```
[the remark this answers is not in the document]
ln-vidrio: The bound parameters `filename` and `path` should probably also be added in if users want to modify a pre-existing chunked dataset. The "open mode" of the datapipe should be noted as a "new" datapipe has different restrictions than with a preexistant one.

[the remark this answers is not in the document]
bendichter: I don't want to change 'data' to 'preload' for a number of reasons, including consistency with pynwb and the fact that this would be confusing if a user just wants to chunk data and does not care about appending. 

How would a user use filename and path?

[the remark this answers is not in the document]
bendichter: Oh, I see now. OK, I can add that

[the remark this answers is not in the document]
ln-vidrio: Regarding `data` <-> `preload`, that's fair. I would still suggest noting that it shouldn't be used after exporting unless that's also part of pynwb convention.

`filename` and `path` essentially make the datapipe readonly save for appending, which I think is important for the user to understand, we can't change the filters after writing.

+types/+untyped/DataPipe.m
bendichter: If no data is provided, `maxSize = [Inf]`? I don't like this because it assumes that the dataset is going to be one-dimensional, which is not a good assumption. Can we make it so that `maxSize` is required if DATA is not provided?

ln-vidrio: Sure, I'll patch it into master.

+types/+untyped/DataPipe.m
```diff
@@ -1,6 +1,47 @@
...
+    %
+    %	DATAPIPE(..., 'data', DATA) Preload DATA in the Dataset. This can
+    %	be omitted if the DATA will be appended later
+    %
+    %   DATAPIPE(..., 'maxSize', MAXSIZE) Sets the maximum size of the HDF5
+    %   Dataset. To append data later, use the MAXSIZE of the full 
+    %   dataset. Inf on any axis will allow the Dataset to grow without
+    %   limit in that dimension. If not provided, MAXSIZE is infered from 
+    %   the DATA. If neither MAXSIZE nor DATA is provided, the assumed
+    %   maximum shape is an infinite vector
```
ln-vidrio, line 16: Just finished patching so this part is invalid.
```suggestion
    %   the DATA. An error is thrown if neither MAXSIZE nor DATA is provided.
```

Issue #231, opened 2020-07-01
Embed schema in NWB file


bahanonu:
Is there a way to embed the NWB schema within the NWB file as it is saved so that the files can be opened in newer versions more seamlessly? `nwbRead` contains the following lines that indicate it is able to read embedded specifications and generate appropriate MATLAB classes: https://github.com/NeurodataWithoutBorders/matnwb/blob/master/nwbRead.m#L39-L47. I am particularly interested in using this for ophys raw and processed data (using `matnwb` api); though I assume any solution should work generally.

For example, I have a file created with NWB `2.0.2` (checked with `util.getSchemaVersion`) but I am running NWB `2.2.5`. Thus, if I try to read the file:
```
nwb = nwbRead('path_to_file.nwb');
```
I get the following error:
```Matlab
Error using types.util.checkUnset (line 13)
Unexpected properties {help}.

Your schema version may be incompatible with the file.  Consider checking the
schema version of the file with `util.getSchemaVersion(filename)` and comparing
with the YAML namespace version present in nwb-schema/core/nwb.namespace.yaml
Error in types.core.ImageSeries (line 38)
            types.util.checkUnset(obj, unique(varargin(1:2:end)));
Error in io.parseGroup (line 85)
    parsed = eval([Type.typename '(kwargs{:})']);
Error in io.parseGroup (line 38)
    subg = io.parseGroup(filename, group, Blacklist);
Error in io.parseGroup (line 38)
    subg = io.parseGroup(filename, group, Blacklist);
Error in nwbRead (line 33)
nwb = io.parseGroup(filename, h5info(filename), Blacklist); 
```

Would want this to be seamless so future users would not have to generate or find the schema that was used to generate that file before reading it.

This appears similar to the issue at https://github.com/NeurodataWithoutBorders/matnwb/issues/224.

ln-vidrio:
Can you please clarify what files you are reading to run into this error?

The error you're seeing is most likely from a file that simply does not actually have an embedded specification. It could also be a MATLAB path conflict where you are referring to previously generated classes.

`util.getSchemaVersion` only checks the `nwb_version` string but it does not check if an embedded spec actually exists in the file itself.

bahanonu:
Thanks, but my question was specifically about embedding the specification/schema, the file is just an example for motivation as to why. Where are we able to let matnwb API know to embed the specification before using `nwbExport`?

[...]

PR #232, opened 2020-07-02
Fix for specification embedding when `namespaces` not in working directory

bahanonu:
Based on discussion in https://github.com/NeurodataWithoutBorders/matnwb/issues/231. I was having inconsistent embedding of the specification when running the `ophys.m` tutorial as a test and during cell extraction in my calciumImagingAnalysis package (https://github.com/bahanonu/calciumImagingAnalysis).

It appears that matnwb including the specification depends on the `namespaces` directory being in the current working directories path, else if you run `schemes.exportJson()` with `namespaces` in another part of the path then you get an empty output and `obj.embedSpecifications` will not embed the specification.

Updated both `schemes.exportJson` and `schemes.loadCache` to find the root matnwb path and load the files in the `namespaces` directory there that are needed to embed the NWB specifications in `nwbExport` created files.

To show this works, I ran the `ophys.m` tutorial in three different conditions. 
- First, I put `namespaces` directory in the current Matlab working directory (root of calciumImagingAnalysis) then ran `nwbExport(nwb, 'ophys_tutorial.nwb');` per the tutorial. Checked using `h5disp('ophys_tutorial.nwb');` and specification was found in `/specifications`.
- Second, I removed the `namespaces` directory from the working directory and re-ran the same code as above and specification was __not__ embedded (`/specifications` empty).
- Third, I ran using the modified code in this pull request with the `namespaces` directory __not__ in the working directory then ran the code as in (1). The specification was embedded.

Likely a more robust solution that will ensure users NWB files are created with embedded specifications. An additional suggested feature would be to warn the user that the specification is not being embedded because the `namespaces` directory and caches could not be found.

### DIFF
--- a/+misc/getMatnwbDir.m
+++ b/+misc/getMatnwbDir.m
@@ -1,0 +1,39 @@
+function [matnwbDir] = getMatnwbDir(varargin)
+	% Find the absolute path location of the matnwb directory.
+	% started: 2020.07.02 [11:53:52]
+	% inputs
+		%
+	% outputs
+		%
+
+	% changelog
+		% 2020.07.02 [11:53:58] - Started function, add warnings. - Biafra Ahanonu
+	% TODO
+		%
+
+	try
+		% Get the actual location of the matnwb directory. This assumes "getMatnwbDir" is within the +misc matnwb package folder.
+		fnLoc = dbstack('-completenames');
+		fnLoc = fnLoc(1).file;
+		[fnDir,~,~] = fileparts(fnLoc);
+		[matnwbDir,~,~] = fileparts(fnDir);
+
+		% Check directory exists else throw a warning letting the user know.
+		dirExists = subfxnDirCheck(matnwbDir,1);
+	catch err
+		disp(repmat('@',1,7))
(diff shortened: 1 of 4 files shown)


ln-vidrio:
@bahanonu Thanks for identifying and fixing the issue! I do have one suggestion if you don't mind moving this code around. Can you move the find namespace functionality and place it in its own function like `misc.getNamespaceDir` or something like that?

I will make a separate pull request later on adding this hierarchy for searches as well as some of the suggested warning messages:
1. Workspace environment
2. Current working directory
3. MatNWB directory
4. Warn that no `namespace` could be found.

Again, thank you for the fixes!

bahanonu:
@ln-vidrio Sure thing! I've made the suggested updates and have it give out a warning message if it cannot find `namespaces` in the `matnwb` folder. Will leave future additions to other pull requests.

Also added a `misc.getMatnwbDir` function (in `+misc`) as this might prove useful going forward to make it so that functions like `generateCore` can run without users needing to change the working directory.

Let me know if you have any questions.

PR #241, opened 2020-11-07
Corrected typo in source code documentation

wuffi:
Calling the function `nwbtests()` would give an error - it should be `nwbtest()`


bendichter:
Thanks @wuffi!

PR #246, opened 2020-12-22
Set up CI with Azure Pipelines

rly:
Add automated run of unit tests using Azure Pipelines. Matlab can be run only on Linux. The configuration yaml specifies the latest version of Matlab but an older version, as old as 2020a can be used.

refs: 
https://www.mathworks.com/help/matlab/matlab_prog/continuous-integration-with-matlab-on-ci-platforms.html
https://marketplace.visualstudio.com/items?itemName=MathWorks.matlab-azure-devops-extension

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,19 @@
+# Starter pipeline
+# Start with a minimal pipeline that you can customize to build and deploy your code.
+# Add steps that build, run tests, deploy, and more:
+# https://aka.ms/yaml
+
+trigger:
+- main
+
+pool:
+  vmImage: 'ubuntu-latest'
+
+steps:
+- script: echo Hello, world!
+  displayName: 'Run a one-line script'
+
+- script: |
+    echo Add other tasks to build, test, and deploy your project.
+    echo See https://aka.ms/yaml
+  displayName: 'Run a multi-line script'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,5 +26,5 @@ steps:
 
   - task: RunMATLABCommand@0
     inputs:
-      command: "try, addpath(genpath(pwd)); generateCore(); nwbtest; catch err, disp(err.message); exit(1); end"
+      command: "try, addpath(genpath(pwd)); generateCore(); results = nwbtest; if isempty(results), exit(1); end; catch err, disp(err.message); exit(1); end"
   
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -15,5 +15,5 @@ steps:
       release: R2020a
   - task: RunMATLABCommand@0
     inputs:
-      command: nwbtest
+      command: "try, addpath(genpath(pwd)); generateCore(); nwbtest; catch err, fprintf('Matlab error: %%s\n', err.message); exit(1); end"
   
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,8 +1,3 @@
-# Starter pipeline
-# Start with a minimal pipeline that you can customize to build and deploy your code.
-# Add steps that build, run tests, deploy, and more:
-# https://aka.ms/yaml
-
 trigger:
 - main
 
@@ -11,8 +6,9 @@ pool:
 
 steps:
   - task: InstallMATLAB@0
-    inputs:
-      release: R2020a
+    # default to latest version of Matlab installed
+    #inputs:
+    #  release: R2020a
 
   - task: UsePythonVersion@0
     inputs:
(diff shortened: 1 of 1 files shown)


bendichter:
woah, cool! Looks like the only tests that are causing problems are the ones that rely on a Python installation with certain dependencies. Good work!

rly:
I added PyNWB to the CI, but now some tests fail because of data type inequality. See: https://dev.azure.com/NeurodataWithoutBorders/matnwb/_build/results?buildId=1299&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=5c7b8cce-d95f-56ce-c8fa-472b8c0934b1

e.g.:
```
================================================================================
Verification failed in tests.system.ElectricalSeriesIOTest/testInFromPyNWB.
    ----------------
    Test Diagnostic:
    ----------------
    Values for property 'data' are not equal
    ---------------------
    Framework Diagnostic:
    ---------------------
    verifyEqual failed.
    --> Classes do not match.
        
        Actual Class:
            int64
        Expected Class:
            int32
    
    Actual Value:
      2x1 int64 column vector
    
       0
       2
    Expected Value:
      2x1 int32 column vector
    
       0
       2
    ------------------
    Stack Information:
    ------------------
    In /home/vsts/work/1/s/+tests/+system/NwbTestInterface.m (NwbTestInterface.verifyContainerEqual) at 68
    In /home/vsts/work/1/s/+tests/+system/NwbTestInterface.m (NwbTestInterface.verifyContainerEqual) at 53
    In /home/vsts/work/1/s/+tests/+system/PyNWBIOTest.m (PyNWBIOTest.testInFromPyNWB) at 28
================================================================================
```

Is this an OS-related issue?

Also, these failures are Verification Failures which do not trigger an error, so CI succeeds even though the tests fail. Should these be assertion tests instead?

ln-vidrio:
This is not an OS issue but DynamicTableRegions are specified as 'int' which I believe we take in as int64. I guess this is no longer the case?

rly:
> This is not an OS issue but DynamicTableRegions are specified as 'int' which I believe we take in as int64. I guess this is no longer the case?

Datasets with dtype 'int' (or 'int32') allow both int32 and int64. The problem seems to be that PyNWB writes some data as int64 but MatNWB writes it as int32 and the unit test expects them to match. 
It does not look like the problem is isolated to 'DynamicTableRegion'. 

Failures are:
tests.system.ElectricalSeriesIOTest/testInFromPyNWB: Values for property 'data' are not equal
tests.system.ElectricalSeriesIOTest/testInFromPyNWB: Values for property 'timestamps_interval' are not equal
tests.system.NWBFileIOTest/testInFromPyNWB: Values for property 'timestamps_interval' are not equal
tests.system.NWBFileIOTest/testInFromPyNWB: Values for property 'data' are not equal
tests.system.NWBFileIOTest/testInFromPyNWB: Values for property 'num' are not equal
tests.system.PhotonSeriesIOTest/testInFromPyNWB: Values for property 'dimension' are not equal
tests.system.PhotonSeriesIOTest/testInFromPyNWB: Values for property 'timestamps_interval' are not equal
tests.system.TimeSeriesIOTest/testInFromPyNWB: Values for property 'timestamps_interval' are not equal
tests.system.ElectricalSeriesIOTest/testInFromPyNWB: Values for property 'timestamps_interval' are not equal

Note that 'data' is 'numeric' in the TimeSeries spec, 'timestamps_interval' is int32 in the TimeSeries spec, 'dimension' is int32 in the ImageSeries spec.

rly:
That said, I think that is a separate issue from the PR at hand. This is ready for review.

ln-vidrio:
I'm not too well-versed with azure pipelines. Is this pipeline run every commit/pull?

rly:
The pipeline is run on every commit to master and every pull request to any branch.

rly:
We can easily configure when it is run. This is how it is set up for PyNWB/HDMF.

ln-vidrio:
That's fine, we should keep it that way honestly.

oruebel:
That's awesome! Having CI running on MatNWB is a major step. Thanks @rly and @In-vidrio!

ln-vidrio:
Whose MATLAB license is being used for the azure pipeline?

rly:
@ln-vidrio No license is needed. Mathworks and Azure/Microsoft must have an agreement / sponsored license. 

> If you want to use a Microsoft-hosted agent, you must include a task in your pipeline to install MATLAB on the agent. Currently, this task is available only for public projects and does not include transformation products, such as MATLAB Coder™ and MATLAB Compiler™.
If you want to instead use a self-hosted agent, you have to install MATLAB on your own machine, presumably with your own license.

Issue #239, opened 2020-09-24
suggestion: make nwb-schema a subrepo of this repo


bendichter:
This would make it easy to check out a specific release tag to read an older file

ln-vidrio:
Since we have customers using the repo links directly, moving to subrepos might break them now unless we find a solution where subrepos are used but are still included in the release asset links.

Issue #17, opened 2018-02-01
Schema should be a submodule


mcafaro:
The schema files have currently been copied and pasted into the MatNWB repository. A better approach may be to use a submodule that points to the schema repo. I can think of some pros/cons for both approaches, so this change is not a definite win but something to think about. Submodule pros: avoids copy/paste mistakes, makes the source of the schema files clear (as well as git rev). Cons: confusing for new users cloning the repo (need to use --recursive flag to get submodules)

ln-vidrio:
Since we now have users that rely on the source code without using git, we can no longer implement this suggested enhancement without breakage.

PR #247, opened 2020-12-23
Schema Selector

### LINKED ISSUE #242: Create a Schema Selector using cached files
For certain files that do not have a cached spec, there needs to be an automated method to generate classes based on a "best guess" based on the nwb_version string.

This method cannot be used with the assumption of a pre-existing `git` command so we will have to cache all previous nwb-schema versions 2.1 and above into their own directories. We would then implement a selector to read from the correct schema version and generate the closest class files from there. Note that this is **only** for cases where the cached schema does not exist in the file itself.

Extensions will still need to be provided by the user.

Tickets that may be in conflict with this change: #239, #17

ln-vidrio:
closes #242, #239, #17

This pr does the following:
- Include nwb schema versions 2.2.x as submodules in the nwb-schema directory
- Implement a version selector in `generateCore` which can be manually selected using a `version` argument with a string indicating the nwb version.
- The previous behavior where `generateCore` could take a list of `namespace.yaml` files and just pipe that to `generateExtension` may be deprecated since one could just use `generateExtension` anyway.
- Add the above behavior to nwbRead if an embedded specification does not exist.


ln-vidrio:
Looks like submodules needs to be added as a step to the azure pipeline.

ln-vidrio:
This change is somewhat breaking so a new release will probably need to be made after merge. That said, I'm pretty sure most users are using the base `generateCore` behavior which has not changed.

PR #253, opened 2020-12-30
Fix Pipeline nwbtest errors

ln-vidrio:


ln-vidrio:
For some reason, generateCore() attempts to create the core.mat cache before the hdmf_common.mat cache despite `core` being depoendent on `hdmf-common`. This does not occur on Windows. Unsure on OSX.

ln-vidrio:
As it turns out, the 2.2.5 hdmf-common submodule wasn't deleted correctly.

Issue #205, opened 2020-04-20
Iterative adding of rows for DynamicTables


bendichter:
Iterative adding of rows to key data structures instead of manual construction.  This includes in particular: `add_trial`, `add_electrode`, `add_unit`. In PyNWB, there is a method, named add_trial that adds trial rows to the trials table one at a time. In MatNWB, all of the data must be supplied for all trials in the initialization of the class. Users like to have the option of specifying data. This should also support `DynamicTable`s that of indexed vectors, like `Units`, allowing you to easily add a vector in an `add_row` update.

Taken from the [Feedback on suggested matNWB feature improvements](https://docs.google.com/document/d/1z7RG_s78a5qsDPTEUVJj5IzsNdclql83Ci3Vb8gd21k) by the Rutishauser Lab

PR #254, opened 2021-01-11
Allow Adding Dynamic Table data by Row.

ln-vidrio:
Dynamic Table (specifically hdmf_common.DynamicTable) and all subclasses now have an `addRow`, `getRow`, and `clear` method.
This allows for adding data to the Dynamic Table by row instead of by column as it is now.

Some notes:
- In order to add rows the `colnames` property must be filled out as that is what is used to verify the shape of the row.
- Empty row data is not possible, which means that all names in `colnames` must contain data. Furthermore, you cannot add a sub-column of a row.
- To allow for Ragged arrays, add a `tablepath` keyword argument that points to the location in the NWB file where your Dynamic Table will be. This only needs to be done the first `addRow` call.
- More specifications for `addRow` can be found in `types.util.dynamictable.addRow` along with the other functions called by the DynamicTable methods.

Known fragility: If the schema moves the following:
* hdmf_common.DynamicTable
* hdmf_common.VectorData
* hdmf_common.VectorIndex
* hdmf_common.ElementIdentifiers
Then this method will most likely break.

fixes https://github.com/NeurodataWithoutBorders/matnwb/issues/205


ln-vidrio:
@bendichter This is relevant to #205 
Anything in particular that is needed other than addRow?

bendichter:
@ln-vidrio looks good to me!

PR #257, opened 2021-01-13
Use RunMATLABTests task in Azure Pipelines

### LINKED ISSUE #255: Tests are failing but CI is passing
Hi,

I just noticed that some tests are failing, for example here:

https://dev.azure.com/NeurodataWithoutBorders/matnwb/_build/results?buildId=1340&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=5c7b8cce-d95f-56ce-c8fa-472b8c0934b1&l=19

but the job is passing. IS this intended. Couple thoughts:

The run MATLAB command line doesn't need its own try-catch. For example this currently in the azure-pipelines file:

`"try, results = nwbtest; if isempty(results), exit(1); end; catch err, disp(err.message); exit(1); end"`

Can be replaced by this:

`"results = assertSuccess(nwbtest);  assert(~isempty(results), 'No tests ran'); "`

The build will fail if the command errors, but the command won't error if a test fails unless you add assertSuccess.

Also, you may consider using the RunMATLABTests task, which also generates Junit and coverage artifacts just like you are doing manually, but may be easier and less error prone. 

Hope that helps!
Andy

rly:
Fix #255.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,20 +21,16 @@ steps:
       pip install pynwb
     displayName: 'Install PyNWB'
 
-  - task: RunMATLABTests@0
+  - task: RunMATLABCommand@0
     inputs:
-      testResultsJUnit: test-results/results.xml
-      codeCoverageCobertura: code-coverage/coverage.xml
-      selectByFolder: +tests
+      command: "results = assertSuccess(nwbtest); assert(~isempty(results), 'No tests ran');"
   - task: PublishTestResults@2
     condition: succeededOrFailed()
     inputs:
-      testResultsFiles: test-results/results.xml
+      testResultsFiles: testResults.xml
   - task: PublishCodeCoverageResults@1
     inputs:
       codeCoverageTool: Cobertura
-      summaryFileLocation: code-coverage/coverage.xml
-  # - task: RunMATLABCommand@0
-  #   inputs:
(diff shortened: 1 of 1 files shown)


ln-vidrio:
Try adding the `selectByFolder` option in `runMATLABtests` as per https://marketplace.visualstudio.com/items?itemName=MathWorks.matlab-azure-devops-extension#run-matlab-tests

rly:
@ln-vidrio Oops! Looks like we made two PRs at the same time (#256).

> Try adding the `selectByFolder` option in `runMATLABtests`
Will do.

ln-vidrio:
@rly I closed mine, yours looks better haha.

ln-vidrio:
I guess you do still need to at least call `generateCore` as a step before running tests just so code coverage can parse `NwbFile.m` correctly.

rly:
Huh, does `nwbtest` call `generateCore`? It seemed to work (at least partially) before...

ln-vidrio:
I think it's something that the task does that's different than `nwbtest`.

I've recently changed the tests to run `generateCore` in temp directories so you shouldn't need to run `generateCore` before running `nwbtest`. I'm not sure why it complains now.

ln-vidrio:
As a side note, when I run the script contents in my local repo, there are no tests found for some reason. I wonder if packages are just ignored.

rly:
I think the pipeline works now. 4 tests are failing due to int32 vs int64 mismatch on a roundtrip with PyNWB (as they were before).

PR #258, opened 2021-01-13
Increase int size when correcting type

ln-vidrio:
Originally used (u)int32 but now uses (u)int64.


ln-vidrio:
That's interesting, the `interval` attribute of `timestamp` in the `TimeSeries` object *should* be an int32.

Issue #221, opened 2020-05-15
Allow loading of non-continuous data segments (DataStub.load)


jens-k:
The method types.untyped.DataStub.load allows addressing a data subset using indices. For example, data.load([23, 2063], [42, 2245]) would give you rows 23 to 42 and columns 2063 to 2245.

However, e.g., when the data represents an LFP, one might be interested in specific non-adjacent rows, i.e. channels. Currently, in order to minimize memory load, one needs to retrieve channel by channel or the first up to the last channel and subselect afterwards. Allowing the user to provide an array with desired channels would feel more natural.

https://nwb-users.slack.com/archives/C5XKC14L9/p1589494856175700

bendichter:
See https://github.com/NeurodataWithoutBorders/matnwb/issues/120

lawrence-mbf:
#259 fixes this issue

PR #259, opened 2021-01-14
Datastub subsref

### LINKED ISSUE #120: use subsref for more intuitive DataStub load
The syntax for data stub load is still a little wonky. Vijay recommended we might think about using [subsref](https://www.mathworks.com/help/matlab/ref/subsref.html) to make the syntax mirror matlab's standard array selection syntax.

ln-vidrio:
Add subsref behavior to DataStub.
This is currently WIP as it may require subtle changes to all classes using DataStub.
Fixes #120, #221

### DIFF
--- a/+tests/+unit/dataStubTest.m
+++ b/+tests/+unit/dataStubTest.m
@@ -21,7 +21,7 @@ nwb = NwbFile(...
     'identifier', 'mouse004_day4', ...
     'session_start_time', session_start_time);
 
-data = reshape(1:5000, 125, 5, 4, 2);
+data = reshape(1:5000, 25, 5, 4, 2, 5);
 
 timeseries = types.core.TimeSeries(...
     'starting_time', 0.0, ... % seconds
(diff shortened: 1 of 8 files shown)


bendichter:
Thanks for working on this!

bendichter:
How do you read the entire dataset? `stub(:,:,:)` should work, but it would also be nice to have a method where you don't need to know the number of dimensions. Does `stub.load` still work?

ln-vidrio:
> How do you read the entire dataset?

`stub(:)` should work as long as the stub is not empty.

`load` was not touched and I've just added a special load_mat_style which just reuses load_h5_style with special conditions.

bendichter:
`(:)` flattens the array though. What is the best approach to read the entire dataset without flattening it?

ln-vidrio:
It actually only returns one element right now, it shouldn't flatten the array.

bendichter:
In standard MATLAB, `(:)` flattens the array. Do you think we should have that subsref act differently here?

ln-vidrio:
Wasn't aware of that, is that true for (:, :) as well for arrays with more than two dimensions?
To load the entire dataset without flattening, you could just require calling `load()`. I'm not sure if you can express that via subsref.

bendichter:
> is that true for (:, :) as well for arrays with more than two dimensions?

I don't think so, I think it's unique to `(:)`

 > To load the entire dataset without flattening, you could just require calling load(). I'm not sure if you can express that via subsref.

That sounds good. I just wanted to make sure that capability remained.

ln-vidrio:
So I tested it out and it actually does flatten based on how many indices there are versus how many dimensions are in the dataset, so I'll be adding that.

bendichter:
Hm ok, I've never seen it used outside of the single dimension case. Sounds good!

ln-vidrio:
@bendichter 
Caveats:
- DataStub is Sealed so no subclassing of DataStub
- DataStub.load() is still the recommended way to load the entire dataset at a normal shape.

PR #269, opened 2021-03-03
Fix JSON exporting issues

### LINKED ISSUE #268: inconsistencies with cached spec
A cached spec with booleans (e.g. an attribute where required=False) is written as required: "false" (false as a string instead of a bool). This writes without error but causes a build error when it is read in pynwb with load_namespaces=True. @owoolnough came across this error and I have not yet been able to reproduce the write error.

ln-vidrio:
fixes #268


bendichter:
@owoolnough can you confirm that this branch works for writing the data in MATLAB and reading it in Python? This would obviate the hot fix I gave you earlier

bendichter:
@owoolnough says this works

PR #290, opened 2021-08-09
Tighten Character Array checks in Attributes

### LINKED ISSUE #289: descriptions should not be allowed to be cell arrays
The "description" field should not be able to be an array. A user can enter

```matlab
timeseries = types.core.TimeSeries( ...
    'data', [1,2,3], ...
    'description', {'aa' 'aa'}, ...
    'starting_time_rate', 100, ...
    'data_unit', 'na');
```

and no error is raised, but this results in an object with an incorrect description that cannot be opened by PyNWB

ln-vidrio:
Fixes #289


bendichter:
@ln-vidrio this works well!

bendichter:
Would it be possible to also add a test so that we can keep our coverage up?

ln-vidrio:
I'll take a look but none of the NwbFile.m coverage is connected to any of the actual changes in the commit.

bendichter:
OK, well no need to hold up the commit if it's not straightforward to do

PR #291, opened 2021-08-10
Release Candidate for Schema 2.4.0

ln-vidrio:


bendichter:
@ln-vidrio I'd like to attempt to create the file created [here](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/gallery/domain/plot_icephys.py) before releasing this new schema.

oruebel:
@bendichter, thanks for adding the tutorial for the new icephys table and testing that it work!

PR #295, opened 2021-08-13
196 direct object reference

### LINKED ISSUE #196: embed path in generated matnwb objects
Currently, in order to create SoftLinks, References, etc. the user needs to know the HDF5 path of the target object. It would be better if the user could just input the object itself rather than a path to that object. The way to solve this would be during export, to take the resolved path of each object and write it as a property within that object. Then we can use that property to resolve links. This would have the added benefit that each object would contain its HDF5 path, which they could use in other software.

ln-vidrio:
Fixes #196


ln-vidrio:
Test notes: the test failures for pynwb are because of the recent changes to schema 2.4.0.

ln-vidrio:
@bendichter Does this work for you? This now adds object references for references and soft links. I intentionally left external links out since there's no guarantee that the file is NWB.

ln-vidrio:
The base commit for this branch is failing so codecov can't provide any details. These fixed tests should at least have the main branch upload codecov files so I can then fix the coverage drops here.

bendichter:
Hey, I am having trouble reading data back with the latest schema:


```matlab
read_nwbfile = nwbRead('ecephys_tutorial1.nwb')

Error using schemes.Namespace/getParent (line 52)
Parent AlignedDynamicTable for class IntracellularRecordingsTable doesn't exist!  Missing Dependency?

Error in schemes.Namespace/getRootBranch (line 95)
            parent = obj.getParent(cursor(obj.TYPEDEF_KEYS{hasTypeDef}));

Error in file.processClass (line 3)
branch = [{namespace.getClass(name)} namespace.getRootBranch(name)];

Error in file.writeNamespace (line 14)
    [processed, classprops, inherited] = file.processClass(className, Namespace, pregenerated);

Error in nwbRead>generateSpec (line 106)
        file.writeNamespace(name);

Error in nwbRead (line 40)
        generateSpec(filename, h5info(filename, specLocation));
```

bendichter:
This is working well. I'm updating the tutorials and utils to match this change

bendichter:
@ln-vidrio looks good to me!

PR #298, opened 2021-08-18
regenerate according to 2.4.0

### LINKED ISSUE #297: API generation tool is dangerously deprecrated
Hey @ln-vidrio, it looks like the [API index](https://neurodatawithoutborders.github.io/matnwb/doc/index.html) is out of date. Would you be able to update it?  How straightforward would it be to maintain old versions of the schema as well?

ln-vidrio:


ln-vidrio:
This will at least update the API docs. I'll spin off the expansion as a separate feature.

PR #302, opened 2021-08-20
Support for Nested VectorIndices

### LINKED ISSUE #301: Bug in Matnwb: use nwb.units.getRow() to retrieve "waveforms" data
I wonder whether you have modified nwb.units.getRow() to get "waveforms" data?
I think I have set "waveforms_index" and "waveforms_index_index" right.
The "waveforms" data can't be retrieved with nwb.units.getRow(), while "spikes_time" can be retrieved normally.

This is what I have:

ln-vidrio:
Fixes #301

### DIFF
--- a/+types/+util/+dynamictable/getRow.m
+++ b/+types/+util/+dynamictable/getRow.m
@@ -31,7 +31,17 @@ end
 
 for i = 1:length(columns)
     cn = columns{i};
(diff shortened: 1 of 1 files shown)


ln-vidrio:
@bendichter This PR was originally meant for #301 but does include important fixes for the case for nested indices which are not supported in the current release. This will most likely be a patch release.

bendichter (approved):

PR #311, opened 2021-09-29
remove positive number validation for getRow ind.

### LINKED ISSUE #310: Errror using dynamicTable.getRow when ID is 0
Hi,
When I ran this,
`row = nwb.units.getRow(0, 'useID', true, 'columns', {'spike_times'});`

I got the following error.

```
Error using types.util.dynamictable.getRow (line 13)
Expected input to be positive.

Error in types.hdmf_common.DynamicTable/getRow (line
135)
        row = types.util.dynamictable.getRow(obj, id,
        varargin{:});
```

This can be hackily solved by commenting out ln.13 in types.util.dynamictable.getRow.

bendichter:
fix #310, #573

If useID is True, then ind can be of value 0

### DIFF
--- a/+tests/+system/DynamicTableTest.m
+++ b/+tests/+system/DynamicTableTest.m
@@ -148,6 +148,20 @@ classdef DynamicTableTest < tests.system.RoundTripTest & tests.system.AmendTest
     end
 
     methods (Test)
+        function getNegativeIdTest(testCase)
+            vector = types.hdmf_common.VectorData('description', 'data column', 'data', (1:10) .');
+            ids = types.hdmf_common.ElementIdentifiers('data', ((-9):0) .');
+            testtable = types.hdmf_common.DynamicTable( ...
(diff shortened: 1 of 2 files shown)


ln-vidrio (changes requested):
Since `ind` isn't used until `useId` is checked for on line 29, I would like to still check that the index is within a valid range if  `useId` is *not* set. Something like:

```MATLAB
if p.Results.useId
    validateattributes(ind, {'numeric'}, {'vector'});
    ind = getIndById(DynamicTable, ind);
else
    validateattributes(ind, {'numeric'}, {'positive', 'vector'});
end
```

bendichter:
@ln-vidrio that sounds fine

lawrence-mbf:
pipeline failure allowable as that should be resolved by #575

lawrence-mbf (approved):

PR #323, opened 2021-10-15
updated types.util.dynamictable.getIndex to properly find VectorIndex target

### LINKED ISSUE #322: ragged column arrays not formatted properly when VectorData object is passed in to ObjectView
Issue with the behavior of ObjectView. In theory, it is possible to pass in the object itself rather than the full hdf5 path to the object. However, this doesn't seem to be the case.

For example, when I run the following code to create a DynamicTable with ragged array columns:
```matlab
    'data', {'1a'; '1b'; '1c'; '2a'} ...
);
col1_index = types.hdmf_common.VectorIndex( ...
    'description', 'column #1 index',...
    'target',types.untyped.ObjectView('/col1'), ...  % object view of target column
    'data', [3; 4] ...
);
table_ragged_col = types.hdmf_common.DynamicTable( ...
    'description', 'an example table', ...
    'colnames', {'col1','col1_index'}, ...
    'col1', col1, ...
    'col1_index', col1_index, ...
    'id', types.hdmf_common.ElementIdentifiers('data', [0; 1]) ...  % 0-indexed, for compatibility with Python
);
table_ragged_col.getRow(1)
```

The output is, as expected:
```
       col1       col1_index
    __________    __________

    {3×1 cell}        3    
```
However, using the suggested method of casting the object itself as an object view:
```matlab
col1_index = types.hdmf_common.VectorIndex( ...
    'description', 'column #1 index',...
    'target',types.untyped.ObjectView(col1), ...  % object view of target column
    'data', [3; 4] ...
);

table_ragged_col = types.hdmf_common.DynamicTable( ...
    'description', 'an example table', ...
    'colnames', {'col1','col1_index'}, ...
    'col1', col1, ...
    'col1_index', col1_index, ...
    'id', types.hdmf_common.ElementIdentifiers('data', [0; 1]) ...  % 0-indexed, for compatibility with Python
);

table_ragged_col.getRow(1)
```
Return the clearly wrong output.
```
     col1     col1_index
    ______    __________

    {'1a'}        3     
```
Clearly, VectorIndex column is not formatting the ragged array column properly in the second scenario.
Let me know if there's something I'm not doing right.

cechava:
resolves #322 

tracked down bug to the following line of code in types.util.dynamictable.getIndex.m
```matlab
tf = VectorIndex.target== DynamicTable.vectordata.get(column);
```
modified it to the following so that the proper properties are compared.
```matlab
tf = VectorIndex.target.target== DynamicTable.vectordata.get(column);
```
Adding the extra .target allows access to the relevant field in the ObjectView object

### DIFF
--- a/+types/+util/+dynamictable/getIndex.m
+++ b/+types/+util/+dynamictable/getIndex.m
@@ -63,7 +63,7 @@ if VectorIndex.target.has_path()
 elseif isprop(DynamicTable, column)
     tf = VectorIndex.target == DynamicTable.(column);
 else
-    tf = VectorIndex.target == DynamicTable.vectordata.get(column);
+    tf = VectorIndex.target.target == DynamicTable.vectordata.get(column);
 end
 end
 


cechava:
Issue is fixed in #302

PR #325, opened 2021-10-19
addition of utility functions to add columns to an already-built DynamicTable object

### LINKED ISSUE #324: addColumn

cechava:
fix #324

### DIFF
--- a/+types/+util/+dynamictable/addVarargColumn.m
+++ b/+types/+util/+dynamictable/addVarargColumn.m
@@ -1,9 +1,12 @@
 function addVarargColumn(DynamicTable, varargin)
 
-%check proper argument/value inputs and parse
-[newVectorData, ~] = types.util.parseConstrained(DynamicTable,'vectordata', 'types.hdmf_common.VectorData', varargin{:});
-
-newColNames = DynamicTable.validate_colnames({varargin{1:2:end}});
+%parse inputs
+p = inputParser();
+p.KeepUnmatched = true;
+p.StructExpand = false;
+parse(p, varargin{:});
+newColNames = DynamicTable.validate_colnames(fieldnames(p.Unmatched));
+newVectorData = p.Unmatched;
 % get current table height - assume id length reflects table height
 if ~isempty(DynamicTable.colnames)
     tableHeight = length(DynamicTable.id.data);
@@ -11,7 +14,7 @@ end
 
 for i = 1:length(newColNames)
     new_cn = newColNames{i};
-    new_cv = newVectorData.get(new_cn);
+    new_cv = newVectorData.(new_cn);    
     % check height match before adding column
     if ~isempty(DynamicTable.colnames)
         indexName = getIndexInSet(newVectorData,new_cn);
@@ -20,22 +23,32 @@ for i = 1:length(newColNames)
                 'NWB:DynamicTable:AddColumn:MissingRows',...
                 'New column length must match length of existing columns ') 
         else
-            assert(height(newVectorData.get(indexName).data) == tableHeight,...
+            assert(height(newVectorData.(indexName).data) == tableHeight,...
                 'NWB:DynamicTable:AddColumn:MissingRows',...
                 'New column length must match length of existing columns ') 
         end
     end
-    if ~isa(new_cv, 'types.hdmf_common.VectorIndex')
-        DynamicTable.colnames{end+1} = new_cn;
+    if 8 == exist('types.hdmf_common.VectorIndex', 'class')
+        if ~isa(new_cv, 'types.hdmf_common.VectorIndex')
+            DynamicTable.colnames{end+1} = new_cn;
+        end
+    else %legacy case
+        if ~isa(new_cv, 'types.core.VectorIndex')
+            DynamicTable.colnames{end+1} = new_cn;
+        end
     end
     DynamicTable.vectordata.set(new_cn, new_cv);   
 end
 end
 
-function indexName = getIndexInSet(inputSet, inputName)
+function indexName = getIndexInSet(inputStruct, inputName)
     % wrap input set with an empty dynamic table
     T = types.hdmf_common.DynamicTable();
-    T.vectordata = inputSet;
+    % convert input structure to a set
+    columnNames = fieldnames(inputStruct);
+    for i = 1:length(columnNames)
+        T.vectordata.set(columnNames{i},inputStruct.(columnNames{i}));
+    end
     % use dynamic table function to get index name
     indexName = types.util.dynamictable.getIndex(T, inputName);
 end


bendichter:
@cechava looks like a good start. I think it's a good time to start adding tests.

+tests/+system/DynamicTableTest.m
```diff
@@ -85,11 +85,38 @@ function appendContainer(testCase, file)
...
+            rag_col = types.hdmf_common.VectorData( ...
+                'description', 'new ragged column', ...
+                                'data', (1000:-1:1) .' ...
+            );
+            rag_col_index = types.hdmf_common.VectorIndex( ...
+                'description', 'new ragged column index',...
+                'target',types.untyped.ObjectView(rag_col), ...
+                'data', sort(randperm(1000,200)) .' ...
+            );
+            % append ragged column
```
bendichter, line 115: you should be able to use `util.create_indexed_column` here

+types/+util/+dynamictable/addColumn.m
```diff
@@ -0,0 +1,41 @@
+function addColumn(DynamicTable, varargin)
+% ADDCOLUMN Given a dynamic table and a set of keyword arguments for one or
+% more columns, add one or more columns to the dynamic table by providing
+% either keywords or a MATLAB table
+%
+%  ADDCOLUMN(DT,table) append the columns of the MATLAB table to the
```
bendichter, line 6:
```suggestion
%  ADDCOLUMN(DT,TABLE) append the columns of the MATLAB Table TABLE to the 
```

bendichter:
This looks good to me. @ln-vidrio , what do you think?

+tests/+system/DynamicTableTest.m
bendichter: Also, this technique won't necessarily get it perfectly because the max value of the index needs to be the length('data') + 1

cechava: because I am using synthetic data, use of `util.create_indexed_column` would require me to break-up the 'data' column vector into individual cells of an array. `util.create_indexed_column` would then figure out the data bounds that make up the 'data' entry in the `VectorIndex` object. It's more straightforward to just define the `VectorIndex` myself since I am synthesizing those bounds.

bendichter: well, it's not more straightforward if it is not correct. The max of rag_col_index needs to be equal to the length of rag_col + 1. Actually, it would be nice if we added that as a check.

cechava: So, not sure if there's a bug somewhere, but in order for the DynamicTable to work properly the last element of rag_col_index needs to be equal to the length of rag_col. (no +1 needed)
For example, the code below works properly:
```matlab
    'data', {'1a'; '1b'; '1c'; '2a'} ...
);

col1_index = types.hdmf_common.VectorIndex( ...
    'description', 'column #1 index', ...
    'target',types.untyped.ObjectView(col1), ...  % object view of target column
    'data', [3; 4] ...
);

table_ragged_col = types.hdmf_common.DynamicTable( ...
    'description', 'an example table', ...
    'colnames', {'col1'}, ...
    'col1', col1, ...
    'col1_index', col1_index, ...
    'id', types.hdmf_common.ElementIdentifiers('data', [0; 1]) ...  % 0-indexed, for compatibility with Python
);

>> table_ragged_col.getRow(2)

ans =

  table

       col1   
    __________

    {1×1 cell}
```

whereas using [3;5] for the 'data' field of the `VectorIndex` object will results in an error when trying to retrieve values from the second row of the table.

```matlab
    'data', {'1a'; '1b'; '1c'; '2a'} ...
);

col1_index = types.hdmf_common.VectorIndex( ...
    'description', 'column #1 index', ...
    'target',types.untyped.ObjectView(col1), ...  % object view of target column
    'data', [3; 5] ...
);

table_ragged_col = types.hdmf_common.DynamicTable( ...
    'description', 'an example table', ...
    'colnames', {'col1'}, ...
    'col1', col1, ...
    'col1_index', col1_index, ...
    'id', types.hdmf_common.ElementIdentifiers('data', [0; 1]) ...  % 0-indexed, for compatibility with Python
);

>> table_ragged_col.getRow(2)
Index in position 1 exceeds array bounds (must not exceed 4).

Error in types.util.dynamictable.getRow>select (line 75)
        selected = Vector.data(selectInd{:});

Error in types.util.dynamictable.getRow>select (line 101)
        selected{iRange} = select(DynamicTable,...

Error in types.util.dynamictable.getRow (line 44)
    row{i} = select(DynamicTable, indexNames, ind);

Error in types.hdmf_common.DynamicTable/getRow (line 121)
        row = types.util.dynamictable.getRow(obj, id, varargin{:});
```

cechava: This is actually also true in the python hdmf (e.g., see ragged array column example [here](https://hdmf.readthedocs.io/en/stable/tutorials/plot_dynamictable_howto.html#dynamictable-howtoguide) so I think the last element of rag_col_index  should indeed be equal to the length of rag_col, as it is here

bendichter: If you use `sort(randperm(1000,200)) .', then the value of the last element is stochastic

cechava: ah, I see what you mean. I will correct to make sure last element is always 1000

+types/+util/+dynamictable/addTableColumn.m
```diff
@@ -0,0 +1,24 @@
...
+if ~isempty(DynamicTable.colnames)
+    tableHeight = length(DynamicTable.id.data);
+end
+
+for i = 1:length(newColNames)
+    new_cn = newColNames{i};
+    new_cv = types.hdmf_common.VectorData( ...
+        'description', 'new column', ...
+        'data', subTable.(new_cn) ...
+        );
```
ln-vidrio, line 15: Should check for the early schema case where VectorData was actually in the `core` namespace similar to how it's done in https://github.com/NeurodataWithoutBorders/matnwb/blob/43258a7d4c6b6b831803d5f4abf64a3b3c3e0518/%2Btypes/%2Butil/%2Bdynamictable/addColumn.m#L25-L29

cechava:
Let me know if it all looks good.

+types/+util/+dynamictable/addVarargColumn.m
```diff
@@ -0,0 +1,42 @@
+function addVarargColumn(DynamicTable, varargin)
+
+%check proper argument/value inputs and parse
+[newVectorData, ~] = types.util.parseConstrained(DynamicTable,'vectordata', 'types.hdmf_common.VectorData', varargin{:});
+
+newColNames = DynamicTable.validate_colnames({varargin{1:2:end}});
```
ln-vidrio, line 6: I would suggest using MATLAB's internal `inputParser` object for this. `parseConstrained` is primarily intended as a generated class helper.

ln-vidrio: Take a look at https://github.com/NeurodataWithoutBorders/matnwb/blob/84123523c4f415a0c394d2bb861f9df6bba463b9/%2Btypes/%2Butil/%2Bdynamictable/addVarargRow.m#L2-L11 for an example.

Normally I would say "if it works it works" but `parseConstrained` and `types.untyped.Set` are subject to change.

ln-vidrio (approved):

PR #331, opened 2021-10-21
deal with empty row IDs when building DynamicTable

### LINKED ISSUE #321: Cannot append MATLAB table to DynamicTable object without id's
When executing the following code:

```matlab
    'data', [1;2] ...
);

col2 = types.hdmf_common.VectorData( ...
    'description', 'column #2', ...
    'data', {'a';'b'} ...
);

my_table = types.hdmf_common.DynamicTable( ...
    'description','an example table', ...
    'colnames', {'col1','col2'}, ...
    'col1',col1, ...
    'col2',col2, ...
    'id', types.hdmf_common.ElementIdentifiers('data', [0;1]) ...  %0-indexed, for compatibility with Python
);

col1 = [4; 5; 6];
col2 = {'d'; 'e'; 'f'};


sub_table = table(col1, col2);
my_table.addRow(sub_table);

```

I get the following MATLAB error pointing to an undefined variable within the addTableRow function :
```
Unrecognized function or variable 't'.

Error in types.util.dynamictable.addTableRow (line 66)
    idRange = (newStartId:(newStartId+height(t))) .';
```

This is not an issue when adding a row by the variable arguments methods as in the following code
```
my_table.addRow('col1', 3, 'col2', {'c'})
```
In this case the underlying function will simply add an index to the new row corresponding  to the row's location index (0-indexed).

cechava:
fixes #321


cechava:
@bendichter 
This PR accomplishes 3 things

1.  If a DynamicTable is created without id's, we create them (done in Python HDMF, but not here and creates problems down the line)
2.  If a DynamicTable is created with columns of different lengths, we throw a useful error (this is done in the Python HDMF, but not here)
3. If rows are added to a DynamicTable with a table and no 'id' column is in that table, we create the necessary ids (this had a bug)

Question:
Change 1 required addition of a line to the types.hdmf_common.DynamicTable.m script. However, this would require changing of the schema since this is a script generated by the generateCore function. Any tips on how to proceed?

cechava:
Nevermind that question. Dug around the schema code and it's clear that it's better not to mess with it. 
Figured out that I could just do the DynamicTable check before addition of any rows. This way the users will also get an error if their DynamicTables have uneven columns. 
**Still a work in progress**. I need to address the case where we're adding rows that are meant to have an 'id_index' column.

PR #333, opened 2021-10-22
NWB dynamic table to MATLAB table

cechava:
should behave like .to_dataframe() method in Python HDMF package


+util/nwb2table.m
```diff
@@ -0,0 +1,76 @@
+function matlabTable = nwb2table(DynamicTable, index)
```
bendichter, line 1: so I originally used this "2" syntax from matlab's num2str, but I think this style has fallen out of style. Maybe just `nwb_to_table` would be better. Better yet would be a `to_table` method on `DynamicTable`

+util/nwb2table.m
```diff
@@ -0,0 +1,76 @@
...
+%NWB2TABLE converts from a NWB DynamicTable to a MATLAB table 
+%
+%   MATLABTABLE = NWB2TABLE(T) converts object T of class types.core.DynamicTable
+%   into a MATLAB Tale
+%   
+%   MATLABTABLE = NWB2TABLE(T, FALSE) includes rows referenced by a
+%   DynamicTableRegion as nested subtables
```
bendichter, line 8:
```suggestion
%   MATLABTABLE = NWB2TABLE(T, INDEX) If INDEX is FALSE, includes rows referenced by a
%   DynamicTableRegion as nested subtables
```

+util/nwb2table.m
cechava: @bendichter I agree, that would be better. There's actually 3 or 4 methods in these PRs that would be better incorporated as methods of the DynamicTable class. Those changes would require changing the schema since that is how the code to define the DynamicTable class is generated. I can look into how to do this and maybe we can queue them up for the next release version

bendichter: Are you sure? How was `addRow` added? That's not anywhere in the schema.

ln-vidrio: @cechava Take a look at Line 99 of `file.fillClass`. You'll have to do something wacky like that for `to_table`

bendichter:
@cechava you can add to https://github.com/NeurodataWithoutBorders/matnwb/blob/655cc50c7a2fd10efbe457675a27afe5cc0e650e/%2Bfile/fillDynamicTableMethods.m  There must be a better way to do this, thought, right? Can't you write a .m file and then read it? Wouldn't that be better than putting it in as a string?

+util/nwb2table.m
cechava: thanks! I wrongly assumed everything in the DynamicTable class definition was added from the schema.

ln-vidrio:
> 
> 
> @cechava you can add to https://github.com/NeurodataWithoutBorders/matnwb/blob/655cc50c7a2fd10efbe457675a27afe5cc0e650e/%2Bfile/fillDynamicTableMethods.m There must be a better way to do this, thought, right? Can't you write a .m file and then read it? Wouldn't that be better than putting it in as a string?

If there is I'm not aware of it. Using a separate file would allow for syntax highlighting but I'm not sure that's a great enough reason considering how simple this snippet actually is. The syntax highlighting would also be potentially misleading since these are object methods and not standalone functions.

If a given generated function body is too complex, it would be better in its own standalone function (like in `types.util.dynamictable.*`).

+file/fillDynamicTableMethods.m
```diff
@@ -8,6 +8,10 @@
     '    row = types.util.dynamictable.getRow(obj, id, varargin{:});',...
     'end',...
     '',...
+    'function table = to_table(obj, varargin)',...
+    '    table = types.util.dynamictable.nwb_to_table(obj, varargin{:});',...
+    'end',...
+    '',...
     'function clear(obj)',...
```
ln-vidrio, line 15: For user interfaces, please stick with camel case for consistency.

+types/+util/+dynamictable/nwb_to_table.m
```diff
@@ -0,0 +1,76 @@
...
+        ref_table = cv.table.target;
+        cv = cell(length(row_idxs),1);
+        for r = 1:length(row_idxs)
+            cv{r,1} = ref_table.getRow(row_idxs(r)+1);
+        end
+        matlabTable.(cn) = cv;
+        columns(i) = [];
+    else
+        i = i+1;
+    end
```
ln-vidrio, line 68: In NWB version `2.0.2` (and maybe later), `DynamicTableRegion` is actually under the `core` namespace which should also be checked for.

ln-vidrio (commented):
https://github.com/NeurodataWithoutBorders/matnwb/pull/333/files#r758541239

https://github.com/NeurodataWithoutBorders/matnwb/pull/333/files#r758544932

ln-vidrio (approved):

PR #335, opened 2021-10-25
dynamic table check

### LINKED ISSUE #334: no DynamicTable check
Currently, there's no check on the shape or attributes of a DynamicTable object either at construction or when adding additional rows. This results in the user potentially creating a table with errors and/or that cannot be exported to an nwb file. 
Every dynamic table must pass two checks: 
1) the length of all columns should be the same
2) the length of row id's should match the length of columns. If the id field is empty, the row indices should be automatically created

This behavior is currently present in the Python HDMF package when constructing a new DynamicTable.

cechava:
fixes #334.

Modified addRow function to check dynamic table shape before adding more rows. Specifically, the following is done: 
1. If a DynamicTable does not contain id's, we create them (done in Python HDMF, but not here and creates problems down the line)
2. If a DynamicTable contains columns of different lengths, we throw a useful error (this is done in the Python HDMF, but not here)

In order to pass DynamicTableTest had to modify addTableRow function to append missing row id's to list rather than use addRawData function which erroneously tries to create a new VectorIndex object within the vectordata set of the dynamic table object. Appending to a list is also the approach taken by the addVargRow function so it should not be a problem.


cechava:
@bendichter looks like the github checks are failing because of lack of test coverage for routine MATLAB functions such as `rand()` and `repmat()` . Any tips on how to deal with this?

cechava:
@ln-vidrio let me know if you have any thoughts/ideas on this pull request. Currently, the check on the dynamic table shape/configuration is only performed when adding a new row.
I am now working to do a check when the `DynamicTable` object is first instantiated. Running into some issues in the case when when the table is part of a NWB file that is read in because the table is gradually constructed which leads to some empty columns and thus raises the 'unmatched column' error. Trying to figure out a way around this.

+types/+util/+dynamictable/addTableRow.m
```diff
@@ -66,7 +66,7 @@ function addTableRow(DynamicTable, subTable, varargin)
     idRange = (newStartId:(newStartId+height(subTable)-1)) .';
 end
 
-types.util.dynamictable.addRawData(DynamicTable, 'id', idRange);
+DynamicTable.id.data = [double(DynamicTable.id.data); idRange];
```
ln-vidrio, line 69: Does this change have something to do with the id types? If so, I believe this is fixed in a different PR no?

+types/+util/+dynamictable/checkConfig.m
```diff
@@ -0,0 +1,86 @@
...
+    ignoreList = {};
+else
+    ignoreList = varargin{1};
+end
+% do not check specified columns - useful for classes that build on DynamicTable class 
+columns = setdiff(DynamicTable.colnames,ignoreList);
+% keep track of last non-ragged column index; to prevent looping over array twice
+c = 1;
+lastStraigthCol = 0;
+lens = zeros(length(columns),1);
```
ln-vidrio, line 25: This is a nit. Please use the full term `lengths` so as not to be confusing with lenses or lensing.

+types/+util/+dynamictable/checkConfig.m
```diff
@@ -0,0 +1,86 @@
...
+if nargin<2
+    ignoreList = {};
+else
+    ignoreList = varargin{1};
+end
+% do not check specified columns - useful for classes that build on DynamicTable class 
+columns = setdiff(DynamicTable.colnames,ignoreList);
+% keep track of last non-ragged column index; to prevent looping over array twice
+c = 1;
+lastStraigthCol = 0;
```
ln-vidrio, line 24: typo `lastStraigthCol` -> `lastStraightCol`

+types/+util/+dynamictable/addTableRow.m
cechava: correct. that was fixed in another PR. this change is no longer necessary. switching it back

ln-vidrio (approved):

PR #342, opened 2021-11-12
modification to types.util.checkDtype  to keep cast integer type

### LINKED ISSUE #339: int type error
It looks like `int`s are being saved in the wrong byte size. The following should produce int16s but it produces int8s.

```matlab
spikes = {[1., 2., 3.], [2., 3., 4., 5.]};

[spike_times_vector, spike_times_index] = util.create_indexed_column(spikes);

nwb = NwbFile( ...
    'session_description', 'mouse in open exploration',...
    'identifier', 'Mouse5_Day3', ...
    'session_start_time', datetime(2018, 4, 25, 2, 30, 3) ...
);

nwb.units = types.core.Units( ...
    'colnames', {'spike_times'}, ...
    'description', 'units table', ...
    'id', types.hdmf_common.ElementIdentifiers( ...
        'data', int16(0:length(spikes) - 1) ...
    ), ...
    'spike_times', spike_times_vector, ...
    'spike_times_index', spike_times_index ...
);

nwbExport(nwb, 'colnames_demo.nwb')
```

image

@ln-vidrio any ideas here would be greatly appreciated.

cechava:
Currently, types.util.checkDtype calls types.util.correctType function when any array is passed in with `value = 'int'`. In the case when the user has cast the array to a specific integer precision type (e.g., int16), we want to respect the class of the array. 
This PR achieves this by skipping over the call to types.util.correctType when the array integer byte size is already specified (i.e., when it's of type int8, int16, int32, or int64).

fix #339

### DIFF
--- a/+types/+util/correctType.m
+++ b/+types/+util/correctType.m
@@ -5,6 +5,8 @@ function val = correctType(val, type)
 
 %check different types and correct
 
+class_val = class(val); %store initial class
+
 if any(strcmp(type, {'logical', 'bool'}))
     assert(islogical(val) || isnumeric(val),...
         'NWB:CorrectType:NonLogical',...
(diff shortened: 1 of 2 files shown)


bendichter:
@ln-vidrio , could you take a look at this?

ln-vidrio:
We could probably just modify `correctType` function directly since a lot of code will get invalidated here. As far as I know, there isn't another use for this function elsewhere.

cechava:
modified with @ln-vidrio's suggestion

ln-vidrio (approved):

bendichter (approved):

+types/+util/correctType.m
```diff
@@ -63,13 +65,15 @@
...
 classMatch = regexp(class(val), 'u?int(\d+)', 'once', 'tokens');
 classSize = str2double(classMatch{1});
-
-intSizeScale = [8 16 32 64];
-for iSize = find(minSize == intSizeScale, 1):find(classSize == intSizeScale, 1)
-    sizeType = sprintf('%s%d', prefix, intSizeScale(iSize));
-    if all(minVal >= double(intmin(sizeType))) && all(maxVal <= double(intmax(sizeType)))
-        val = cast(val, sizeType);
-        break;
+if ~any(strcmpi(class_val, {'int8' 'int16' 'int32' 'int64'}))
```
bendichter, line 68: shouldn't we do this for uint as well?

ln-vidrio: That would cover all integral types, maybe `~isinteger` would be more concise if that's what we want?
https://www.mathworks.com/help/matlab/ref/isinteger.html

PR #346, opened 2021-11-17
Associate cell strings with simple spaces.

### LINKED ISSUE #338: cell array with single value saved as scalar
When you save a cell array with a single element that is a string attribute, e.g. `{'spike_times'}`, this is saved as a scalar, which causes a validation error with `pynwb.validate`.

This bug has appeared in `DynamicTable.colnames`, which should be an array of strings. If only a single column is provided, this value is converted to a scalar string and causes a validation error.

This is preventing us from uploading some data generated from MatNWB to DANDI.

```matlab
spikes = {[1., 2., 3.], [2., 3., 4., 5.]};

[spike_times_vector, spike_times_index] = util.create_indexed_column(spikes);

nwb = NwbFile( ...
    'session_description', 'mouse in open exploration',...
    'identifier', 'Mouse5_Day3', ...
    'session_start_time', datetime(2018, 4, 25, 2, 30, 3) ...
);

nwb.units = types.core.Units( ...
    'colnames', {'spike_times'}, ...
    'description', 'units table', ...
    'id', types.hdmf_common.ElementIdentifiers( ...
        'data', int16(0:length(spikes) - 1) ...
    ), ...
    'spike_times', spike_times_vector, ...
    'spike_times_index', spike_times_index ...
);

nwbExport(nwb, 'colnames_demo.nwb')
```

```python
from pynwb import validate, NWBHDF5IO

io = NWBHDF5IO('/Users/bendichter/dev/matnwb/colnames_demo.nwb','r')
validate(io)
```
```
[Units/colnames (units.colnames): incorrect shape - expected an array of shape '[None]', got non-array data 'spike_times',
 ElementIdentifiers (units/id): incorrect type - expected 'int', got 'int8',
 Units/id (units/id): incorrect type - expected 'int', got 'int8']
```

@ln-vidrio do you think you could direct us to where in the code we should look to fix this?

ln-vidrio:
Fixes #338

### DIFF
--- a/+io/mapData2H5.m
+++ b/+io/mapData2H5.m
@@ -23,7 +23,7 @@ if ischar(data)
         end
         sid = H5S.create_simple(1, size(data,1), max_dims);
     end
-elseif ~forceArray && isscalar(data)
+elseif ~forceArray && ~iscell(data) && isscalar(data)
     sid = H5S.create('H5S_SCALAR');
 else
     if isvector(data)
--- a/+io/parseAttributes.m
+++ b/+io/parseAttributes.m
@@ -47,8 +47,6 @@ for i=1:length(attributes)
         H5T.close(tid);
         H5A.close(aid);
         H5F.close(fid);
-    elseif isscalar(attr.Value) && ~iscell(attr.Value)
-        args(attr.Name) = attr.Value{1};
     else
         args(attr.Name) = attr.Value;
     end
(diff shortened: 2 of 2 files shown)


cechava (commented):
looks good. issue is fixed. should we be concerned about the coverage test fail?

ln-vidrio:
Normally yes but it looks like it would only remove this one line. Not great but as long as it's above 85% I think we're fine.

https://github.com/NeurodataWithoutBorders/matnwb/blob/f710fde6b9557caefdce70d6d751bd7d612b72e3/NwbFile.m#L72-L74

PR #371, opened 2021-12-22
Refactor dim checks and reformat error message

### LINKED ISSUE #370: better shape constraint error message
see https://github.com/NeurodataWithoutBorders/helpdesk/discussions/16

ln-vidrio:
Fixes #370 

The dimension checks were a bit too loose and would have required further validation
by external tools. This tightens them to be more conforming.

Reformatted the dimension check error message as the constraints used by nwb might not
be the most clear.


+types/+util/checkDims.m
```diff
@@ -1,54 +1,57 @@
...
+    expectedRank = length(expectedSize);
 
-if isVector
-    if isscalar(adjsz)
-        flipsz = [1 adjsz];
+    if 1 == expectedRank
+        % since MATLAB doesn't actually support single ranks,
+        % we must check the ambiguous case where the vector might be
+        % vertical or horizontal. This orientation doesn't actually matter
+        % at the HDF5 level.
```
bendichter, line 18: I don't think this true. I think the mapping should be:
```
MATLAB         HDF5
=======       =======
[1, N]         [N, ]
[N, 1]         [1, N]
impossible     [N, 1]
```

ln-vidrio: Huh, maybe that was changed.

bendichter: Well wait a minute what do you think it is/should be?

ln-vidrio: Oh wait, I think it actually does coerce vectors. See these lines here:
https://github.com/NeurodataWithoutBorders/matnwb/blob/223a0c47d711d4948064d28e7f39c966a09bac1f/%2Bio/mapData2H5.m#L29-L35

We force vectors to be contiguous regardless of their orientation on write.

ln-vidrio: I think this is primarily UX for MATLAB since horizontal vertices are (anecdotally) used more often than vertical ones (spaces are easier than semicolons).

bendichter: Yes, I agree the UX for MATLAB is an issue. So how would you change the table above?

ln-vidrio: | MATLAB    |   HDF5 |
| ------------ | ----------  |
| [1, N]         | [N]       |
| [N, 1]        | [N]        |
| impossible | [1,N]    |
| impossible | [N,1]    |

This is what it is currently I think. If we do have [N, 1] or [1, N] we'll have to consider changing this.

bendichter: is this a recent change or what we always had?

ln-vidrio: I really can't recall, it's what I reach for intuitively though so I'd imagine it's just how I've done things.

bendichter: looping in @cechava because we were talking about this yesterday.

My confusion was that the table above applies to DataPipe objects, but not to standard matrices.

bendichter: I suppose we could say if you want to make a (1,N) array you have to use a `DataPipe`

bendichter: there are conceivable times when you would want to make a (1,N) array, for instance if you had a table that had a vector column that had only one row.

ln-vidrio: Do you mean a dynamic table or a MATLAB table? I would not expect either to have such an issue.

bendichter: I mean what if we had a column that is of shape [N,K] in HDF5, with N as the number of rows of the table. In MATLAB, that would be [K,N]. But what if you just wanted to add a single row? That would be [1,K] in HDF5 and [K,1] in MATLAB. The DataPipe object currently supports writing vectors like this and does not automatically change them to [K,] in HDF5.

PR #378, opened 2022-01-04
updated getRow function to slice through last dimension of multidimensional columns

### LINKED ISSUE #377: getRow slices wrong dimension for multidimensional matrices
for a DynamicTable with a column of shape `(P,O,M,N)`, the _getRow_ method should slice through the last dimension, which specifies the column height for multidimensional columns.
Currently, the _getRow_ method slices through the first dimension regardless of matrix dimensionality.

cechava:
fixes #377


cechava:
with initial commit, getRow method works with simple DynamicTable objects

sample snippet
```matlab
    'data', (1:10)' ...
);
col2 = types.hdmf_common.VectorData( ...
    'description', 'column #2', ...
    'data', rand(5,2,10) ...
);
my_table = types.hdmf_common.DynamicTable( ...
    'description','an example table', ...
    'colnames', {'col1','col2'}, ...
    'col1',col1, ...
    'col2',col2, ...
    'id', types.hdmf_common.ElementIdentifiers('data', (0:9)') ...
);
>> my_table.getRow(1:3)

ans =

  3×2 table

    col1         col2     
    ____    ______________

     1      [1×5×2 double]
     2      [1×5×2 double]
     3      [1×5×2 double]
```

bendichter:
Could we make the rows 5x2 instead of 1x5x2?

cechava:
> Could we make the rows 5x2 instead of 1x5x2?

it's not immediately clear how to do that. That output is handled under the hood by MATLAB after running the following line
```matlab
subTable = table(row{:}, 'VariableNames', columns);
```
with the following variable values
```matlab
K>> row

row =

  1×2 cell array

    {3×1 double}    {3×5×2 double} 
    
K>> columns

columns =

  1×2 cell array

    {'col1'}    {'col2'} 
```

+tests/+system/DynamicTableTest.m
```diff
@@ -6,37 +6,53 @@ function addContainer(~, file)
             for c =1:length(colnames)
                 colnames{c} = char([double(colnames{c}) zeros(1,randi(10))]);
             end
+            % until addRow works with multidimensional matrices, define array in one go
+            nrows = 200;
+            ids = primes(2000)';
```
bendichter, line 11: Why are you generating the first 2000 prime numbers? I don't see how that has anything to do with what you are testing

cechava: It actually generates all prime numbers up to 2000. This was there before and doesn't hinder any test so I thought I would keep it. From what I gather, prime numbers are used as ids to have a very different way to enumerate rows to make sure the the 'useID' argument of getRow works properly. We could also do something simpler like enumerating rows in intervals of 5, if you prefer.

ln-vidrio (approved):

bendichter (approved):

PR #379, opened 2022-01-04
Deprecation of MATLAB table as input to addRow and addColumn DynamicTable methods

cechava:
calling `addTableRow` or `addTableColumn` functions will throw an error instructing users to use key-value pairs instead


+tests/+sanity/GenerationTest.m
```diff
@@ -38,18 +38,21 @@ function dynamicTableMethodsTest(testCase)
                 'colnames', colnames);
             
             id = primes(2000) .';
-            for i = 1:100
+            for i = 1:200
```
bendichter, line 41: why generate so many random variables? It seems like this test would work just as well with 5 rows.

ln-vidrio (approved):

cechava:
@bendichter if this also looks good to you. As requested, I reduced the number of random numbers generated for the tests

+tests/+sanity/GenerationTest.m
```diff
@@ -37,21 +37,24 @@ function dynamicTableMethodsTest(testCase)
                 'description', 'test dynamic table column',...
                 'colnames', colnames);
             
-            id = primes(2000) .';
-            for i = 1:100
+            id = primes(200) .';
```
bendichter, line 40: why are you generating primes though?

bendichter: again, you are using `.'` in your syntax and I do not understand why. This appears to do the same thing as `primes(200)`'

cechava: Commented on this on PR #378 as well. Copying here:
This was there before and doesn't hinder any test so I thought I would keep it. From what I gather, prime numbers are used as ids to have a very different way to enumerate rows to make sure the the 'useID' argument of getRow works properly. We could also do something simpler like enumerating rows in intervals of 5, if you prefer.

cechava: I can remove the dot. As you point out, it does nothing in this context.

ln-vidrio: So I've been using the `.'` transpose style. It usually does nothing but I wanted to be clear on what I'm doing to the matrix itself. The difference between `transpose` and `ctranspose` here:
https://www.mathworks.com/help/matlab/ref/transpose.html
https://www.mathworks.com/help/matlab/ref/ctranspose.html

ln-vidrio: The usage of `primes` was also me and it's just as @cechava says I just needed a non-linear id vector. As long as it's not the default `id` vector structure it should be testing the same thing. That's just there to differentiate row index and row IDs.

bendichter: @ln-vidrio thanks for pointing me to that reference. I was unaware of the difference between transpose and ctranspose. I don't think we really deal with complex values in NWB, but I can see why you'd default to `.'` nonetheless. Looks good to me!

bendichter (approved):

+tests/+sanity/GenerationTest.m
cechava: @ln-vidrio yes, for pointing to that.

PR #382, opened 2022-01-07
expand docs to show how MATLAB dataset dimensions map onto HDF5 dataset dimensions

cechava:


README.md
```diff
@@ -85,23 +85,44 @@ NWB files use the HDF5 format to store data. Due to differences in how MATLAB an
 
 [without DataPipes](https://neurodatawithoutborders.github.io/matnwb/tutorials/html/dimensionMapNoDataPipes.html)
 
+**Writing to File**
+
 | Shape <br /> in MatNWB| Shape<br />in HDF5|
 | :----------: | :----------: |
 |    (N, 1)    |     (N,)     |
 |    (1, N)    |     (N,)     |
 | (P, M, O, N) | (N, O, M, P) |
```
bendichter, line 94: I feel like N,M,O,P is a more intuitive order

README.md
```diff
@@ -79,12 +79,55 @@ PyNWB's cached schemas are also supported, bypassing the need to run `generateCo
 
 MatNWB is available online at https://github.com/NeurodataWithoutBorders/matnwb
 
+## Data Dimensions
+
+NWB files use the HDF5 format to store data. Due to differences in how MATLAB and HDF5 represent data, the dimensions of datasets are flipped when writing to/from file in MatNWB. This behavior differs depending on whether ```VectorData``` use ```DataPipe``` objects to contain the data. It's important to keep in mind the mappings below to make sure is written to and read from file as expected.
```
bendichter, line 84:
```suggestion
NWB files use the HDF5 format to store data. There are two main differences between the way MATLAB and HDF5 represents dimensions. The first is that HDF5 is C-ordered, which means it stores data is a rows-first pattern, and the MATLAB is F-ordered, storing data in the reverse pattern, with the last dimension of the array stored consecutively. The result is that the data in HDF5 is effectively the transpose of the array in MATLAB. The second difference is that HDF5 can store 1-D arrays, but in MATLAB the lowest dimensionality of an array is 2-D. Due to differences in how MATLAB and HDF5 represent data, the dimensions of datasets are flipped when writing to/from file in MatNWB. This behavior differs depending on whether ```VectorData``` use ```DataPipe``` objects to contain the data. It's important to keep in mind the mappings below to make sure is written to and read from file as expected.
```

README.md
```diff
@@ -79,12 +79,56 @@ PyNWB's cached schemas are also supported, bypassing the need to run `generateCo
...
+|    (N, 1)    |     (N,)     |
+|    (1, N)    |     (N,)     |
+| (P, O, M, N) | (N, M, O, P) |
+
+**Reading from File**
+
+| Shape <br /> in HDF5| Shape<br />in MatNWB|
+| :----------: | :----------: |
+|     (N,)     |     (N,1)    |
+| (N, M, O, P) | (P, O, M, N) |
```
bendichter, line 101: a bit odd to not use alphabetical order

cechava: I was trying to be consistent by having N always denote the number of rows. I will change to be in alphabetical order. For consistency, M will be the # of rows in the table.

bendichter (approved):

PR #385, opened 2022-01-07
restore check for DynamicTable shape

cechava:


+tests/+system/DynamicTableTest.m
```diff
@@ -61,40 +64,40 @@ function addExpandableDynamicTable(~, file, start_array, stop_array, ...
             start_time_exp = types.hdmf_common.VectorData( ...
                 'description', 'start times', ...
                 'data', types.untyped.DataPipe( ...
-                    'data', start_array', ...
-                    'maxSize', [Inf, 1], ...
-                    'axis', 1 ...
+                    'data', start_array, ...
+                    'maxSize', [1, Inf], ...
+                    'axis', 2 ...
```
bendichter, line 69: Can we make this 1-D?

cechava: yes. I can change that once PR #386 is merged so that the tests don't fail.

bendichter (approved):

Issue #388, opened 2022-01-13
Vector indices don't support DataPipes


cechava:
this is necessary to create expandable tables with ragged arrays.

code to reproduce:
```matlab
random_exp = types.hdmf_common.VectorData( ...
    'description', 'random data column', ...
    'data', types.untyped.DataPipe( ...
        'data', rand(5,1), ...  #random data
        'maxSize', [Inf, 1], ...
        'axis', 1 ...
    ) ...
);
random_exp_index = types.hdmf_common.VectorIndex( ...
    'description', 'index to random data column', ...
     'target',types.untyped.ObjectView(random_exp), ...
    'data', types.untyped.DataPipe( ...
        'data', [5], ...  
        'maxSize', [Inf, 1], ...
        'axis', 1 ...
    ) ...
);

Operands to the || and && operators must be convertible to logical scalar
values.

Error in types.util.correctType (line 39)
        assert(0 <= minVal && maxVal <= double(intmax('uint64')),...

Error in types.util.checkDtype (line 111)
            val = types.util.correctType(val, type);

Error in types.hdmf_common.VectorIndex/validate_data (line 36)
        val = types.util.checkDtype('data', 'uint8', val);

Error in types.hdmf_common.Data/set.data (line 31)
        obj.data = obj.validate_data(val);

Error in types.hdmf_common.Data (line 24)
        obj.data = p.Results.data;

Error in types.hdmf_common.VectorData (line 15)
        obj = obj@types.hdmf_common.Data(varargin{:});

Error in types.hdmf_common.VectorIndex (line 15)
        obj = obj@types.hdmf_common.VectorData(varargin{:});
```

Issue is with assert statement within types.util.correctType. Currently, it can't handle empty arrays. Should be a quick fix

cechava:
closed with PR #389

PR #389, opened 2022-01-13
amend types.util.correctType to support empty arrays

cechava:
fixes issue #388


cechava:
current test errors will disappear with fixes from PR #385

ln-vidrio (approved):

PR #396, opened 2022-02-16
checkConfig now skips empty colnames

### LINKED ISSUE #394: CheckConfig should allow "simply empty" colnames in Dynamic Tables
Currently will throw an error since the type is technically not a string type.

Certain types like AlignedDynamicTable does allow empty `colnames` which is usually stored as an empty integer.

lawrence-mbf:
Fixes #394


ln-vidrio:
@lvsltz This should allow you to read those difficult icephys files you have.

lvsltz:
Indeed the file generated from the [pynwb tutorial](https://pynwb.readthedocs.io/en/stable/tutorials/domain/plot_icephys.html#imports-used-in-the-tutorial) is correctly read, but not the other which had the DataPipe error. I'll open a different issue for this one.
Thanks!

PR #401, opened 2022-02-17
Expand squeezed expected size in checkDims

### LINKED ISSUE #399: checkDims does not account for MATLAB squeezing scalar dimensions
In the case of `/000054_Giocomo/sub-R4/sub-R4_ses-20190824T210000` the `checkDims` function does not account for MATLAB auto-squeezing out dimensions beyond 2 if they're all 1.

lawrence-mbf:
Fixes #399


bendichter:
lgtm. Once we confirm that this fixes the problem I think we should merge :-)

PR #419, opened 2022-04-26
add tutorial for setting up dynamically loaded filters

bendichter:


lawrence-mbf:
@bendichter The OSX/Linux mention should be placed above the instructions IMO. I would consider just dropping the MATLAB documentation for Windows users instead of opening the command prompt.

Can't do suggestions since `mlx` is technically a zipped binary I think.

bendichter:
why is this different for Windows?

bendichter:
@lawrence-mbf how's this

Issue #417, opened 2022-03-11
Move docs to use Sphinx


lawrence-mbf:
- [ ] Regular docs
- [ ] use [matlabdomain](https://github.com/sphinx-contrib/matlabdomain) to auto-generate class files.

lawrence-mbf:
related to #297

PR #425, opened 2022-05-03
Propose logo for MatNWB

oruebel:
## Motivation

We now have logos for most NWB-related tools (PyNWB, HDMF, NWBInsepctor etc.) and I figured it would be good to also have a logo for MatNWB. Since the Matlab logo is copyright protected I figured it would not be wise to try and do some form of combination of the NWB and Matlab logos. The logo I'm proposing here places the NWB logo in euclidean coordinate frame to represent both the Mat and NWB part of the name. The PowerPoint file has both a 2D and a 3D variant of the logo, but I think the 3D version looks more interesting. What do you think?

For now, I have place the logo into a new top-level folder. I was not sure where the right place for this is. Just let me know in case I should move the files to a different location. 

## Checklist

- [X] Have you ensured the PR description clearly describes the problem and solutions?
- [X] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/neurodatawithoutborders/matnwb/pulls) for the same change?
- [X] If this PR fixes an issue, is the first line of the PR description `fix #XX` where `XX` is the issue number?


bendichter:
cute!

lawrence-mbf:
My gut says that the logo should be under `docs` but we may want to do that after #417 is done.

lawrence-mbf (approved):

oruebel:
> the logo should be under `docs` but we may want to do that after #417 is done.

Sounds good. I added an item to #417 to note that we should consider moving the logos folder, just so we don't forget.

PR #435, opened 2022-06-07
Support 2.5.0

lawrence-mbf:
## Motivation

Support NWB Schema 2.5.0

## How to test the behavior?
Run `nwbtest` with source files without (new) errors.

## Checklist


rly:
FYI NWB 2.5.0 has been released.

lawrence-mbf:
@rly @oruebel It looks like there is still an error introduced with 2.5.0. In the schema, what does the `default_value` indicate if the value is also optional (required: false)? Right now we're automatically setting `data_offset` in the `TimeSeries` object to `0.0` since that's the default value but it appears to be empty in PyNWB.

rly:
@lawrence-mbf Sorry I am just seeing this. If the field is optional (`required: false`), then the `default_value` represents what the API and users should interpret the value to be if the field is not present. It looks like this is no longer an issue though with the latest PyNWB. Is that correct?

rly:
With MatNWB now supporting NWB schema 2.5.0, should there be a new release?

lawrence-mbf:
@rly Yes, this must have been fixed, thus I merged it during the User days. I'll see to drafting a release in a bit.

PR #464, opened 2022-09-07
Check for R2017a or older when using jsonencode

lawrence-mbf:
## Motivation

`jsonencode` doesn't support the keyword arguments used to convert inf or NaN values to null in json, making conversion loops impossible for these versions.

The fixes now will throw a warning on write.

## How to test the behavior?
```MATLAB
nwbExport(nwb);
```
Should be successful on MATLAB R2017b and newer and a warning otherwise.

## Checklist


bendichter (approved):
looks good to me

PR #534, opened 2023-08-30
add new raw/scratch script

lawrence-mbf:
## Motivation

Adds Raw/Scratch tutorial from [this pynwb guide](https://pynwb.readthedocs.io/en/stable/tutorials/general/scratch.html)

- [x] Add HDFView image confirming written external link
- [x] Add link to tutorial in README
- [x] refactor multiline function format
- [x] Run files through NWB Inspector

## How to test the behavior?
Documentation should be correct and cover the same topics

## Checklist


bendichter:
@lawrence-mbf thanks for putting this together!

1. I changed the style from

```matlab
func( ...
    'arg', arg ...
   , 'arg2', arg2');
```

to
 ```matlab
func( ...
    'arg', arg, ...
    'arg2', arg2');
```

For the tutorials, I am adopting a style that is an approximate translation from Python's Black. I appreciate that your approach provides the benefit that removing a line only changes that one line, which accounts for the fact that a trailing is allows in Python but not in MATLAB. However, I have never seen code in this style and I think it might be confusing. Are you following an established style standard here? Otherwise, I'd prefer to bite the bullet and put the commas on the line of the arg, even though I appreciate the point that we are losing the benefit of having line changes be completely self-contained.

2. Could you add a screenshot of HDFView showing that the resulting file does indeed use an external link?

3. Is it necessary to have the scratch demo also use external links? It seems like this is combining two concepts that do not need to be combined and making the tutorial more complex than it needs to be. If a user is interested in using the scratch space, I'd prefer to have a tutorial that does not require them to understand external links first. Maybe it would make sense to separate these into completely different mlx files?

4. Could you please add a link from the README to the html files similar to we have for the other tutorials?

**Comment by lawrence-mbf** [comment]
@bendichter 

## (1) Regarding multiline function style:
AFAIK there is no coding standard in MATLAB other than auto-indentation (CTRL+I). This adoption is purely from modifying MATLAB's name-value args. I'm not sure why adopting a python-specific style would help MATLAB style but I can change that for the script anyway.

## (3) External Links
This was pulled from the documentation [here](https://pynwb.readthedocs.io/en/stable/tutorials/general/scratch.html) and I'm trying to follow the major topic as closely as I can. It is true that external linkage is a viable solution for miscellaneous data usage so I wanted to make that clear that you can do that too (albeit, manually).

**Comment by lawrence-mbf** [comment]
@bendichter @oruebel updated with suggestions

File: tutorials/html/scratch.html [hunk 1]
```diff
@@ -0,0 +1,165 @@
...
+)
+ProcModule = types.core.ProcessingModule( ...
+    'description', 'a module to store filtering results', ...
+    'filtered_timeseries', FilteredTs ...
+);
+ContextFile.processing.set('core', ProcModule);
+nwbExport(ContextFile, 'context_file.nwb');
+%% Warning Regarding the Usage of Scratch Space
+% *Scratch data written into the scratch space should not be intended for reuse 
+% or sharing. Standard NWB:N types, along with any extensions, should always be
```
rly, line 121:
```suggestion
% or sharing. Standard NWB types, along with any extensions, should always be 
```

tutorials/html/scratch.html
```diff
@@ -0,0 +1,165 @@
...
+<br>
+<!-- 
+##### SOURCE BEGIN #####
+%% Scratch Data
+% This tutorial will focus on the basics of working with a <https://neurodatawithoutborders.github.io/matnwb/doc/+types/+core/NWBFile.html 
+% |NWBFile|> for storing non-standardizable data. For example, you may want to 
+% store results from one-off analyses of some temporary utility. NWB:N provides
```
rly, line 70:
```suggestion
% store results from one-off analyses of some temporary utility. NWB provides 
```

PR #556, opened 2023-12-20
Pregenerate class files in the repo

### LINKED ISSUE #555: [Feature]: do we still need to require user to `generateCore();`?

@lawrence-mbf, this is a question that @rly brought up in our last meeting and I am also curious.

Do we still need to require users to call `generateCore()` to install MatNWB? What if we pre-generated the classes for the latest schema? This would make the installation process simpler.

_No response_

See above

Yes.

### Code of Conduct

lawrence-mbf:
## Motivation

Fixes #555

## How to test the behavior?
cloning the repo should not require `generateCore` when accessing classes.

## Checklist


bendichter:
Looks great. Let's also adjust the tutorials, since generating the classes is no longer necessary

lawrence-mbf:
@bendichter I think I got most of them. There's one that is kind of necessary in the "Reading NWB Data in MATLAB" tutorial since that does cover what happens if you read a file that does not have an embedded schema.

tutorials/html/intro.html
```diff
@@ -201,31 +207,28 @@
...
 %% Installing MatNWB
 % Use the code below within the brackets to install MatNWB from source. MatNWB 
-% works by automatically creating API classes based on the schema. Use <https://neurodatawithoutborders.github.io/matnwb/doc/generateCore.html 
-% |*generateCore*|> to generate these classes.
+% works by automatically creating API classes based on the schema.
 
 %{
 !git clone https://github.com/NeurodataWithoutBorders/matnwb.git
 cd matnwb
-addpath(genpath(pwd));
```
bendichter, line 227: Is this line OK to delete?

tutorials/html/intro.html
```diff
@@ -385,7 +388,7 @@
 %% Read
 % We can then read the file back in using MatNWB and inspect its contents. 
 
-read_nwbfile = nwbRead('intro_tutorial.nwb')
+read_nwbfile = nwbRead('intro_tutorial.nwb', 'ignorecache')
```
bendichter, line 391: why ignore cache here?

tutorials/html/intro.html
lawrence-mbf: Actually, it should be `addpath(genpath(pwd))` and `cd matnwb` should probably be removed instead.

tutorials/html/intro.html
lawrence-mbf: The environment is already generated (we wrote it in the section above) so we don't need to re-generate the cache. Furthermore, this preserves tests that run this tutorial.

lawrence-mbf: To clarify, since we do need to switch between schema versions in the tutorials, we generate the cache locally in a working directory before cleaning it up, ensuring that the test environment is empty before we run the next test (which could be using any environment). As a result, I believe all tests and tutorials should be using `ignorecache` and its variants, otherwise you will end up with strange class reference errors.

bendichter (approved):

Issue #262, opened 2021-02-03
matnwb package


bahanonu:
Is there any plan to move the current `matnwb` packages into sub-packages of a `matnwb` package for a cleaner namespace? 

For example, instead of calling `types.core.OpticalChannel` users could call `matnwb.types.core.OpticalChannel` and this would be true for all functions in the current top-level packages. 

This would also reduce potential conflicts given the names of the current top-level packages—whereas currently, `matnwb` may have to be added or remove from the path each time a NWB read/write operation is performed to avoid this (e.g. if a parent function has `types= '';` or similar command, issues could arise).

This would not be a minor change (since other packages and code depend on the current setup), but may be worth considering.

ln-vidrio:
No but your timing is pristine as I can say that the alternative `table` constructor in native MATLAB does break for precisely this namespace reason. Evidently, the script used for the internal `tabular` class uses a variable called `types` which interferes with regular MatNWB use, especially when allocating space using ObjectViews.

You are correct that this will be quite a bit of effort as not only does the code regularly reference other functions within itself, the generated class locations will all need to be changed, and there will definitely be breakage in the utility scripts and other third-party applications that utilize the generated classes and custom types.

bendichter:
I like namespaces as much as the next Python programmer, but this seems like a lot.

bahanonu:
@ln-vidrio 
Thanks for the heads up about `table`, that's a good example case.

It is possible depending on the timescale of implementation, users could be notified of the change (as MathWorks currently does) and that new functions would be put under the `matnwb` package (with several ways that would allow a gradual transition of existing functions to a `matnwb` package). 

For existing users, if a `matnwb` package is made, many would likely just need to start functions/scripts with `import matnwb.*` if they want to use the current naming conventions.

Else can just rely on adding/removing matnwb from path as needed to avoid these issues.

ehennestad:
I support this!

FYI:

Matlab provides an option to define an alias file, which supports backwards compatibility of names. At least 

The [matlab.alias.AliasFileManager](https://se.mathworks.com/help/matlab/ref/matlab.alias.aliasfilemanager-class.html) class provides an API for creating and maintaining alias definitions. Aliases are not part of class definitions. Instead, alias definition files are stored in resources folders that are located in the same folder as the latest class file.

This would ensure that utility functions and third party applications would continue to work as is (for a transition period, lets say).

ehennestad:
[...]

PR #560, opened 2024-02-05
Fix inherited readonly props

ehennestad:
## Motivation
Some schema attributes are redefined to be readonly for inherited classes. In MATLAB a property attribute can not be redefined for subclasses, so it was possible to set readonly properties.

## How to test the behavior?
```
es = types.core.ElectricalSeries()
es.data_unit = 'volt' % works
es.data_unit = 'millivolts' % should not work
```

## Checklist


lawrence-mbf:
@ehennestad This is not kosher but could you apply the fix mentioned in here: https://github.com/NeurodataWithoutBorders/matnwb/issues/262#issuecomment-1923619681 as most of the errors in these tests come from there.

ehennestad:
The failed test is actually a good sign!

lawrence-mbf (dismissed):
lgtm

lawrence-mbf (approved):

Issue #572, opened 2024-06-21
[Bug]: Pixel_mask doesn't seem to be properly constructed. 


chrisvdt:
### What happened?

 Although my nwb object does get exported to an NWB file without errors, and the validation with python is successfull, I cannot view the Imagesegmentation in neurosift.

A raw view shows that the pixel_map is not a compound datatype object array.
![afbeelding](https://github.com/NeurodataWithoutBorders/matnwb/assets/5311644/c5334262-4a1c-4b02-ada5-2a51106e8ce1)

I made this structure array with struct('x', uint32(0), 'y', uint32(0), 'weight', single(0)). But I see that matnwb writeCompound converts them all to doubles.

### Steps to Reproduce

```matlab
mask = struct('x', uint32(0), 'y', uint32(0), 'weight', single(0));

                       mask_index = uint8(zeros(nRois, 1));
                       bi = 0;
                       for i = 1:nRois
                           [y, x] = find(ROI_data.Mask == i);
                           ln = length(x);
                                                     
                           mask_index(i) = ln;
                          
                           for j = 1:ln
                                mask(bi + j).x = x(j);
                                mask(bi + j).y = y(j);
                                mask(bi + j).weight = 1.0;
                           end
                           bi = bi + ln;
                       end

                       pix_mask = types.hdmf_common.VectorData('data', mask, 'description', 'roi pixel position (x,y) and pixel weight');
```

### Operating System

Windows

### Matlab Version

matlab R2021a

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/neurodatawithoutborders/matnwb/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] Have you ensured this bug was not already [reported](https://github.com/NeurodataWithoutBorders/matnwb/issues)?

chrisvdt:
I wrote earlier that writeCompound converted all the fields to doubles. That is not true, my structure arrays were not formatted properly. I changed my code and now the structure arrays keep their original data types uint32, uint32, single.
The result is however the same, pixel_mask is an {Object Object] array, and neurosift does not display the pixel_mask.

bendichter:
What it looks like for a file generated with PyNWB:

![image](https://github.com/NeurodataWithoutBorders/matnwb/assets/844306/c9636eba-feb5-4ebd-8048-f9dabe542d13)

ehennestad:
[...]

PR #577, opened 2024-07-01
Add example for creating pixel mask to ophys tutorial (inspired by #572)

ehennestad:
## Motivation

It was unclear how to create pixel_mask for PlaneSegmentation objects. This was raised in issue #572 

## How to test the behavior?
```
open ophys.mlx
```

## Checklist


ehennestad:
@bendichter 
@lawrence-mbf one test fails, related to nwb version. Is that due to matnwb still being on v2.6.0?

lawrence-mbf:
> @bendichter @lawrence-mbf one test fails, related to nwb version. Is that due to matnwb still being on v2.6.0?

Yeah, the message implies that pynwb is using 2.7.0 by default.

tutorials/html/ophys.html
```diff
@@ -301,21 +299,61 @@
...
+    end
+    pixel_mask_index = uint16(cumsum(num_pixels_per_roi)); % Note: Use an integer 
+    % type that can accommodate the maximum value of the cumulative sum
+
+    % Create pixel_mask_index vector
+    pixel_mask_index = types.hdmf_common.VectorIndex(...
+            'description', 'Index into pixel_mask VectorData', ...
+            'data', pixel_mask_index, ...
+            'target', types.untyped.ObjectView(...
+                '/processing/ophys/ImageSegmentation/PlaneSegmentation/pixel_mask') );
```
bendichter, line 326: This should work, but there is a better way. The current approach would require adjustments if the user wanted to provide a different name for the `ImageSegmentation` or `PlaneSegmentation` object. You should be able to first create the pixel_mask `VectorData` object and then put that object in here instead of this internal path. I prefer would that approach because it does not require the user to input the path of the PlaneSegmentation object, and therefore is robust to name changes and a bit more modular.

tutorials/html/ophys.html
```diff
@@ -301,21 +299,64 @@
...
+    % Create pixel_mask_index vector
+    pixel_mask_index = types.hdmf_common.VectorIndex(...
+            'description', 'Index into pixel_mask VectorData', ...
+            'data', pixel_mask_index, ...
+            'target', types.untyped.ObjectView(pixel_mask) );
+
+    plane_segmentation = types.core.PlaneSegmentation( ...
+        'colnames', {'pixel_mask'}, ...
+        'description', 'roi pixel position (x,y) and pixel weight', ...
+        'id', types.hdmf_common.ElementIdentifiers('data', int64(0:n_rois-1)'), ...
```
bendichter, line 336:
```suggestion
```

this line is no longer needed. If `'id'` is omitted, it will be added automatically.

bendichter:
@ehennestad this looks great! Just one minor edit, then it's good to go

ehennestad:
@bendichter

PR #602, opened 2024-10-16
Update load_mat_style.m

### LINKED ISSUE #601: [Bug]: New matnwb not reading files
### What happened?

I upgraded matlab and downloaded the latest mantnwb (2.7.0). I added the path ($HOME/matnwb). When I type 
w=nwbRead('myfyle.nwb) I get this error. I have tried using gerneateExtensions for 2.7.0 and 2.6.0 (my files in older version), but maybe I'm doing this incorrectly . I use this to generateExtensions
generateExtension('$HOME/matnwb/nwb-schema/2.7.0/core/nwb.namespace.yaml').  Am I doing something wrong?

### Steps to Reproduce

```matlab
git clone https://github.com/NeurodataWithoutBorders/matnwb.git
In matlab
addpath()
w=nwbRead(myfile)
```

### Error Message

```matlab
Error using ind2sub (line 27)
Size vector must be a real vector with at least 2 elements.

Error in types.untyped.DataStub/load_mat_style (line 49)
        [points{:}] = ind2sub(dataDimensions, orderedSelection);
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in types.untyped.DataStub/load (line 124)
                data = obj.load_mat_style(varargin{1});
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in indexing (line 157)
                B = builtin('subsref', obj, S);
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in types.util.checkDtype>unwrapValue (line 162)
            unwrapped = wrapped.load(1);
                        ^^^^^^^^^^^^^^^
Error in types.util.checkDtype (line 123)
    value = unwrapValue(value);
            ^^^^^^^^^^^^^^^^^^
Error in types.core.NWBFile/validate_file_create_date (line 335)
        val = types.util.checkDtype('file_create_date', 'datetime', val);
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in types.core.NWBFile/set.file_create_date (line 187)
        obj.file_create_date = obj.validate_file_create_date(val);
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in types.core.NWBFile (line 127)
        obj.file_create_date = p.Results.file_create_date;
        ^^^^^^^^^^^^^^^^^^^^
Error in NwbFile (line 16)
            obj = obj@types.core.NWBFile(varargin{:});
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in io.parseGroup (line 76)
            parsed = NwbFile(kwargs{:});
                     ^^^^^^^^^^^^^^^^^^
Error in nwbRead (line 79)
    nwb = io.parseGroup(filename, h5info(filename), Blacklist);
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### Operating System

Linux

### Matlab Version

2024b

### Code of Conduct

- [X] Have you ensured this bug was not already [reported](https://github.com/NeurodataWithoutBorders/matnwb/issues)?

ehennestad:
Fix #601 

## Motivation

Fix issue which arise when loading 1D vectors in MATLAB R2024b due to changes to the `ind2sub` function as described here:
https://se.mathworks.com/help/releases/R2024b/matlab/ref/ind2sub.html?searchPort=49569#mw_463053e0-3bc5-40de-88c3-5887f667b5d3

## How to test the behavior?
```
Loading NWB-file with a file-create date should not produce error in MATLAB R2024
```

## Checklist


+types/+untyped/@DataStub/load_mat_style.m
```diff
@@ -46,6 +46,17 @@
...
+
+        if isscalar(dataDimensions) && ~isMATLABReleaseOlderThan('R2024b')
+            % Starting from MATLAB R2024, the input argument for the size 
+            % of an array must be a vector of positive integers with two or 
+            % more elements. This fix replicates the behavior of older 
+            % MATLAB versions, which automatically assumed that the size 
+            % referred to the number of rows. For scalar dimensions 
+            % (i.e., row or column vectors), this should be acceptable.
+            dataDimensions = [dataDimensions, 1];
+        end
```
bendichter, line 58: It seems like this fix should continue to work for earlier versions of matlab, right?

ehennestad: This fix is only needed for MATLAB R2024b or newer releases. For older releases no change is needed

ehennestad: isMATLABReleaseOlderThan was introduced in R2020b so this can cause errors for older releases. Will add another condition to ensure it does not fail in such cases

bendichter: Yeah I understand it's only needed for new releases but it should also work fine for older releases, right?

+types/+untyped/@DataStub/load_mat_style.m
```diff
@@ -46,6 +46,22 @@
...
+                % Starting in MATLAB R2024b, the input argument for the size 
+                % of an array in ind2sub must be a vector of positive integers 
+                % with two or more elements. This fix replicates the behavior of 
+                % older MATLAB versions, where it was assumed that the a scalar
+                % size referred to the row dimension. For scalar dimensions 
+                % (i.e., row or column vectors), we can still assume this
+                % to be true in matnwb.
+                dataDimensions = [dataDimensions, 1];
+            end
+        end
```
ehennestad, line 63: I updated the code, hope it is clearer now.

+types/+untyped/@DataStub/load_mat_style.m
```diff
@@ -187,4 +203,12 @@
...
         indexKeyIndex((indexKeyIndexNextIndex+1):end) = 1;
     end
+end
+
+function tf = isMATLABReleaseOlderThanR2024b()
+    if exist('isMATLABReleaseOlderThan', 'file') == 2
+        tf = isMATLABReleaseOlderThan('R2024b');
+    else
+        tf = true; % isMATLABReleaseOlderThan was introduced in R2020b
+    end
```
ehennestad, line 213: This ensures the release check does not fail if run from a release predating R2020b

+types/+untyped/@DataStub/load_mat_style.m
ehennestad: @ben, see updates below

+types/+untyped/@DataStub/load_mat_style.m
ehennestad: https://github.com/NeurodataWithoutBorders/matnwb/blob/89055c2d3779e2633fcdf7ad598c792ba0e97a54/%2Btypes/%2Buntyped/%40DataStub/load_mat_style.m#L65

This is the continuation where dataDimension after R2024b can not be a scalar value anymore

bendichter: My point is, we can just change it without checking the version number. This new way should work for older versions as well. Then we don't need to worry about the isMATLABReleaseOlderThanR2024b() function

+types/+untyped/@DataStub/load_mat_style.m
bendichter: this feels pretty hacky, I'd rather just avoid using this function

+types/+untyped/@DataStub/load_mat_style.m
```diff
@@ -48,18 +48,14 @@
...
-            end
+            % Starting in MATLAB R2024b, the input argument for the size 
+            % of an array in ind2sub must be a vector of positive integers 
+            % with two or more elements. This fix replicates the behavior of 
+            % older MATLAB versions, where it was assumed that the a scalar
+            % size referred to the row dimension. For scalar dimensions 
+            % (i.e., row or column vectors), we can still assume this
+            % to be true in matnwb.
+            dataDimensions = [dataDimensions, 1];
         end
```
ehennestad, line 59: I see what you are saying, yes this is a simpler solution! Good observation

+types/+untyped/@DataStub/load_mat_style.m
```diff
@@ -46,6 +46,18 @@
...
+            % Starting in MATLAB R2024b, the input argument for the size 
+            % of an array in ind2sub must be a vector of positive integers 
+            % with two or more elements. This fix replicates the behavior of 
+            % older MATLAB versions, where it was assumed that the a scalar
+            % size referred to the row dimension. For scalar dimensions 
+            % (i.e., row or column vectors), we can still assume this
+            % to be true in matnwb.
+            dataDimensions = [dataDimensions, 1];
+        end
+
```
ehennestad, line 60: I see what you are saying, yes this is a simpler solution! Good observation

ehennestad, line 60: I see what you are saying, yes this is a simpler solution! Good observation

bendichter (approved):

Issue #603, opened 2024-10-08
[Bug]: File generated with MATNWB 2.4.0 issue 


SPMMER:
### What happened?

We created our NWB files with MATNWB 2.4.0 and now we try to validate it with NWBInspector. It trows the following error:

### Steps to Reproduce

```python
(NWBinspector) PS C:\Users\Stefan\PycharmProjects\NWBinspector> nwbinspector C:\Users\Stefan\Desktop\NWBData_Tes
t\
C:\Users\Stefan\.conda\envs\NWBinspector\lib\site-packages\hdmf\spec\namespace.py:535: UserWarning: Ignoring cac
hed namespace 'hdmf-common' version 1.5.0 because version 1.8.0 is already loaded.
  warn("Ignoring cached namespace '%s' version %s because version %s is already loaded."
C:\Users\Stefan\.conda\envs\NWBinspector\lib\site-packages\hdmf\spec\namespace.py:535: UserWarning: Ignoring cac
hed namespace 'core' version 2.4.0 because version 2.7.0 is already loaded.
  warn("Ignoring cached namespace '%s' version %s because version %s is already loaded."
C:\Users\Stefan\.conda\envs\NWBinspector\lib\site-packages\hdmf\spec\namespace.py:535: UserWarning: Ignoring cac
hed namespace 'hdmf-experimental' version 0.1.0 because version 0.5.0 is already loaded.
  warn("Ignoring cached namespace '%s' version %s because version %s is already loaded."
Traceback (most recent call last):
  File "C:\Users\Stefan\.conda\envs\NWBinspector\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\Stefan\.conda\envs\NWBinspector\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\Users\Stefan\.conda\envs\NWBinspector\Scripts\nwbinspector.exe\__main__.py", line 7, in <module>     
  File "C:\Users\Stefan\.conda\envs\NWBinspector\lib\site-packages\click\core.py", line 1157, in __call__       
    return self.main(*args, **kwargs)
  File "C:\Users\Stefan\.conda\envs\NWBinspector\lib\site-packages\click\core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "C:\Users\Stefan\.conda\envs\NWBinspector\lib\site-packages\click\core.py", line 1434, in invoke
[...]

PR #615, opened 2024-11-04
Change: Only embed specs/namespaces for types that are included in NWB file on export

### LINKED ISSUE #607: [Feature]: Improve embedding of namespaces on nwbExport

All extension namespaces in the cache (namespaces folder), will get embedded in the specs of an NWB file when running nwbExport. This can lead to files containing embedded specifications that are not relevant to any of the data types in the file.

One recorded problem of this in #603

Implement a system detecting what namespaces a user actually used and embedding schemas off of that.

Yes.

### Code of Conduct

ehennestad:
Fix #607 

## Motivation

Ensure that only relevant namespaces/specifications are embedded in an NWB file. This [issue](https://github.com/NeurodataWithoutBorders/matnwb/issues/603#issuecomment-2420917269) suggests that such extra namespaces can cause errors when validating with NWBInspector.

## How to test the behavior?
```
Export NWB file with types from a subset of available extension and check the file's `specification` group
```

## Questions / todo

- [x] Are namespace names always specified in `kebab-case`? - Appears so
- [x] Does matnwb always format namespace names as `snake_case`? - Appears so
- [x] ~~Should the `specification` group be completely cleared before rewriting it if exporting to an existing file?~~
- [x] Check superclasses of each type and included namespace of superclasses as well
- [x] Add unit test

## Checklist


+io/+internal/+h5/deleteGroup.m
```diff
@@ -0,0 +1,16 @@
+function deleteGroup(fileReference, groupLocation)
+% deleteGroup - Delete the specified group from an NWB file
```
bendichter, line 2: I don't think it's a big deal that way this is used here, but note that deleting groups and datasets from an HDF5 file does not free up the space

bendichter:
There are lots of new functions here that lack unit tests

ehennestad:
> There are lots of new functions here that lack unit tests

@bendichter Added unit tests for those functions

bendichter:
This seems quite complicated. Can you describe your general strategy here? I would think we could keep track of the namespaces used as we add neurodata objects to an NWB file and then only save the file with those namespaces. It looks here like you are doing a lot of manipulation on the HDF5 level. Can you help me understand this approach before I dig in?

ehennestad:
> This seems quite complicated. Can you describe your general strategy here? I would think we could keep track of the namespaces used as we add neurodata objects to an NWB file and then only save the file with those namespaces. It looks here like you are doing a lot of manipulation on the HDF5 level. Can you help me understand this approach before I dig in?

Please see the above comment/review.

Most of this PR is convenience functions. I added them to the `io.internal.h5` namespace in anticipation of adding Barr-functionality down the road

ehennestad:
@bendichter I updated the test.

bendichter (approved):

PR #616, opened 2024-11-04
Change display of "empty" types.untyped.Set object (Issue #561)

### LINKED ISSUE #561: [Feature]: Override isempty for types.untyped.Set

At the moment an instance of types.untyped.Set can appear to be empty, but using the `isempty` function returns false. 

```
Example:

>> F = types.core.Fluorescence();
>> F.roiresponseseries

ans = 

  Empty Set

>> isempty(F.roiresponseseries)

ans =

  logical

   0

>> 
```

Override the isempty for types.untyped.Set

```
        % overloads isemty
        function tf = isempty(obj)
            tf = obj.Count == 0;
        end
```

Yes.

### Code of Conduct

ehennestad:
Fix #561 
## Motivation

See issue #561

This changes how a Set with no elements is displayed and updates the size method to return a 1x2 vector to conform with the default behavior of MATLAB's `size` function

## How to test the behavior?
```matlab

% Before
>> A = types.untyped.Set()

A = 

  Empty Set

>> isempty(A)

ans =

  logical

   0

>> size(A)

ans =

  uint64

   0

>> 


% After 
>> A = types.untyped.Set()

A = 

  Set with no elements.

>> isempty(A)

ans =

  logical

   0

>> size(A)

ans =

  1×2 uint64 row vector

   0   1

>> 

```

## Checklist


ehennestad:
The following snippet causes a lot of tests to fail, but is not relevant anymore:

https://github.com/NeurodataWithoutBorders/matnwb/blob/3222f33258775c1836c0aee670b3be1c1f74531b/%2Bio/parseGroup.m#L66-L70

The code was embedded in this context:
```matlab
    %immediately elide prefix all property names with this but only if there are
    %no typed objects in it.
    propnames = keys(props);
    if hasTypes
        parsed = containers.Map({root}, {props});
    else
        parsed = containers.Map;
        for i=1:length(propnames)
            pnm = propnames{i};
            p = props(pnm);
            parsed([root '_' pnm]) = p;
        end
    end
    
    if isempty(parsed)
        %special case where a directory is simply empty.  Return itself but
        %empty
        parsed(root) = [];
    end
```

It was possible to create a containers.Map object named `parsed` that was empty, but `parsed` was later changed to be a `types.untyped.Set` which per definition (ref this PR) can not be empty (unless specifically created with `types.untyped.Set.empty`)

https://github.com/NeurodataWithoutBorders/matnwb/blob/5aa27d92cba1b9136612cc8ab80d0c1791975997/%2Bio/parseGroup.m#L56-L63

Code coverage also shows that lines are not covered by tests. Will remove this part

ehennestad:
Current thinking: This should be not be merged as is. A `types.untyped.Set` is similar to a `struct`, `containers.Map` or `dictionary` in that it holds name-value pairs. Non of the builtin MATLAB types are empty if they have no name-value pairs, i.e `s = struct` is a scalar, and `s = struct.empty` is empty. 

The same should be true for the Set. 

Instead of redefining is empty, I think it would be better to not override size as is currently done, and also to change the custom display for a Set with no elements from:
`Empty Set`
to:
`Set with no elements` or `Set with no name, value pairs`

bendichter:
I'm not really clear on what problem this is solving

ehennestad:
The current behavior is confusing. The following code suggest that the set (A) is empty, but `isempty(A)` returns false.

```matlab
>> A = types.untyped.Set()

A = 

  Empty Set
```

PR #617, opened 2024-11-05
Add ErrorID and warningID for all user-facing code

### LINKED ISSUE #11: Warnings and errors should have IDs
Assigning IDs to warnings makes it possible to suppress them without suppressing all warnings. Assigning IDs to errors makes it easier to distinguish between errors thrown by the code. Not having error IDs, I cannot easily test, for instance, if a particular action causes a particular error.

Some of the warnings and errors in the code already have something that looks like an ID, but its in the message:
```matlab
error('Group:subsasgn: please specify whether this numeric value is in ''attributes'' or ''datasets''')
```
These should be pulled out of the message and assigned to the message ID:
```matlab
error('Group:subsasgn',  'Please specify whether this numeric value is in ''attributes'' or ''datasets''')
```
Or a warning/error ID naming strategy should be defined for the project.

ehennestad:
Fix #11 

## Motivation
Fix issue

## How to test the behavior?
```
N/A
```

## Checklist


ehennestad:
Todo:

- [ ] Go through all calls to the `assert` function.

bendichter (approved):

PR #634, opened 2024-11-21
Update link validation function (issue #559)

### LINKED ISSUE #559: [Bug]: Softlinks does not enforce the correct type, nor is it required to add a linked type as a softlink
### What happened?

This issue related to the following points:
1. I can add a linked type directly to an NWB type/class without using the `types.untyped.Softlink`.
Example (adding a device):
```
% Create a device
device = types.core.Device();
nwb.general_devices.set('Device', device);

imaging_plane = types.core.ImagingPlane( ...
    'description', 'a very interesting part of the brain', ...
    'device', device, ...
```

2. I can add another type to the `device` property using a SoftLink class.

Example:
```
% Create an optical channel.
optical_channel_green = types.core.OpticalChannel( ...
    'description', 'green', ...
    'emission_lambda', 500.);
    
 imaging_plane = types.core.ImagingPlane( ...
    'description', 'a very interesting part of the brain', ...
    'device', types.untyped.SoftLink(optical_channel_green), ...
```

In both cases the nwb export works fine.

This issue might be related to this one:
https://github.com/NeurodataWithoutBorders/matnwb/issues/48

The concept of h5 links / softlinks might be too specific for the average user to care about so maybe it would be better to handle this internally? 

I.e in the set method for a linked type, if the value is not a `types.untyped.SoftLink` convert it to one? This would take away the burden from the user to check whether a type is linked or "embedded" (i.e an optical channel is "embedded", not linked as far as I can see). It would also ensure that it's harder to add types that are not correct, as in point 2.

Further, if the user provides an object of type  `types.untyped.SoftLink`,  the validator should check that the `target` of the `types.untyped.SoftLink` is of the correct type.
 

See here for three .m scripts to reproduce:
https://www.dropbox.com/scl/fo/bdvm4tgc79n2w4sn4f2ij/h?rlkey=wccijdyeaxm1b5gonzwmcu0w2&dl=0

### Steps to Reproduce

```matlab
See linked files above
```

### Error Message

```matlab
No errors
```

### Operating System

macOS

### Matlab Version

MATLAB Version: 23.2.0.2409890 (R2023b) Update 3
MATLAB License Number: 885740
Operating System: macOS  Version: 13.5 Build: 22G74 
Java Version: Java 1.8.0_392-b08 with Amazon.com Inc. OpenJDK 64-Bit Server VM mixed mode

### Code of Conduct

- [X] Have you ensured this bug was not already [reported](https://github.com/NeurodataWithoutBorders/matnwb/issues)?

ehennestad:
Fix #559 

## Motivation
Fix issue

## How to test the behavior?
```
See issue for examples
```

## Checklist


ehennestad:
Todo:
Add better error message if the wrong target type is given for a property value which is a link

bendichter (approved):

bendichter:
To do:

* update tutorials to not use explicit softlinks
* add warning for using string in SoftLink

PR #640, opened 2024-11-27
Fix: Make inherited read-only datasets from schema read-only properties in matnwb

ehennestad:
## Motivation

The `IZeroClampSeries` has [three datasets](https://github.com/NeurodataWithoutBorders/nwb-schema/blob/54f4980e3be54f2c05582371d25b76074e160d40/core/nwb.icephys.yaml#L77-L89) (`capacitance_compensation`, `bridge_balance`, `bias_current`) where the value should be fixed according to the schema.

This fix ensure the values are set to zero when objects are created, and also issues a "read-only" error if users try the set another property value

## How to test the behavior?
```
types.core.IZeroClampSeries()
```
The values of the datasets listed above should be 0, instead of empty ([]) as before

```
types.core.IZeroClampSeries('bias_current', 1)
```
Should now throw an error

## Checklist


ehennestad:
Thanks to @stephprince for helping figuring this out

+types/+core/IZeroClampSeries.m
```diff
@@ -33,58 +33,25 @@
...
-            end
-        elseif istable(val)
-            valsz = [height(val) 1];
-        elseif ischar(val)
-            valsz = [size(val, 1) 1];
+        if isequal(val, 0)
+            val = 0;
         else
-            valsz = size(val);
+            error('NWB:Type:ReadOnlyProperty', 'Unable to set the ''bias_current'' property of class ''<a href="matlab:doc types.core.IZeroClampSeries">IZeroClampSeries</a>'' because it is read-only.')
```
bendichter, line 39: I don't know if "read-only" is quite right. That would imply that it has already been written and is being read, which isn't quite true. How about "because it is a fixed value"

ehennestad: I am following the convention In MATLAB where properties that are Constant or has SetAccess = private/protected (i.e., users cannot set the value), are "read-only".

ehennestad: So with this classdef:
```
classdef SimpleClass
    % SimpleClass - A class with one public and one private property
    
    properties (Constant)
        Name=2
    end
    methods
        % Constructor
        function obj = SimpleClass()
        end
    end
end
```
You will have this error if you try to set the Name property:
```
>> s=SimpleClass();
>> s.Name=3
Unable to set the 'Name' property of class 'SimpleClass' because it is read-only.
```

bendichter: ah, ok. I'm not crazy about that language but it's fine if it's consistent with existing MATLAB usage

bendichter (approved):

bendichter:
looks like there are some test issues from the schema update but the changes from this PR look good

PR #643, opened 2024-12-06
Update documentation to use sphinx

ehennestad:
## Motivation

Previous documentation relied on outdated scripts. This PR adds tools for autogenerating documentation using sphinx-matlabdomain. 

New documentation page: https://matnwb.readthedocs.io/en/latest/index.html

## Todo

- [x] Move the [Data Dimensions](https://github.com/NeurodataWithoutBorders/matnwb?tab=readme-ov-file#data-dimensions) section to a page in the RTD
- [x] Modify the README page to point users to the RTD.
- [x] Improve the height of the iframe.
- [x] The tutorials contain links to the old API docs, and will need to be update to the new API docs
- [ ] Favicon - added, not working in safari
- [x] Include links to the youtube video walkthrough

## Checklist


bendichter:
This looks really great! The tutorials and API docs both look really nice. I love the cross-links in the API docs to the schema and schema docs for each of the classes.

I think the main remaining point here would be to modify the README.md. We shoud:
* Move the [Data Dimensions](https://github.com/NeurodataWithoutBorders/matnwb?tab=readme-ov-file#data-dimensions) section to a page in the RTD
* Modify the README page to point users to the RTD.


I have a few very minor suggestions that should not be blocking:
* Improve the height of the iframe. Currently, the tutorials are displayed in the sphinx page as an iframe. The way it is currently done, the height of the iframe matches the height of the window. This results in two scroll wheels, one for the outer page and one for the iframe. It would be a bit better if the iframe height were increased to match the height of the content, so you would only be left with the outer scroll wheel.
* It would be great to add a the matnwb logo as a favicon instead of using the default RTD favicon.

Next steps:
* The tutorials contain links to the old API docs, and will need to be update to the new API docs

ehennestad:
> which points to the old links. At this point, honestly we could just remove this section. I think it's pretty clear how to navigate to the other tutorials and the relevant documentation using the current RTD structure.

It might be nice to keep these links as convenience for people exploring the livescripts in MATLAB.
I think the best option is to make the links relative (to other tutorials at least). This way, when people follow the link in MATLAB it will open the live script instead of the browser, whereas if people browse it online, it will open the the html page in the browser.

It seems MATLAB's html-export automatically changes the extension of links from .mlx to .html so this should work without interventions long as the relative locations are kept consistent (I did not test it yet, but will try)

bendichter:
Looks awesome! Let's merge as soon as the tests pass

bendichter:
@ehennestad do you want to remove the draft status?

ehennestad:
I want to add a section about the dimensions and how text in the documentation about time series dimensions are inconsistent with how matnwb works

ehennestad:
@bendichter Also this: https://github.com/NeurodataWithoutBorders/nwb-overview/issues/180

bendichter (dismissed):

bendichter (approved):

Issue #650, opened 2025-01-21
[Bug]: File written with MatNWB fails to open with PyNWB


arnodelorme:
### What happened?

This file saved using the NwbFile object

https://nemar-dev.ucsd.edu/dataexplorer/download?filepath=/expanse/projects/nemar/openneuro/nwb_example_files/sub-005_ieeg.nwb

Can be reimported in MATLAB using the same object. However, it fails to be reimported in Python. @dungscout96, you used the two different Python libraries right? Can you explain further?

### Steps to Reproduce

```matlab
Try importing the file in Python.
```

### Error Message

```matlab

```

### Operating System

macOS

### Matlab Version

2024b

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/neurodatawithoutborders/matnwb/blob/master/.github/CODE_OF_CONDUCT.md)
- [x] Have you ensured this bug was not already [reported](https://github.com/NeurodataWithoutBorders/matnwb/issues)?

ehennestad:
Hi,

I have tried to read the file with both MatNWB and PyNWB (v2.8.1), and both worked. What version of PyNWB do you use?
(Use `import pynwb; pynwb.__version__` to check)

arnodelorme:
@dungscout96 can you comment?

arnodelorme:
Oh this is actually this file https://testspeedeegdash.s3.us-east-2.amazonaws.com/sub-002_task-FaceRecognition_eeg.nwb

dungscout96:
Here's the error I'm seeing using pynwb:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
File ~/Documents/EEGLAB/EEG-Dash-Data/.venv/lib/python3.11/site-packages/hdmf/build/objectmapper.py:1369, in ObjectMapper.construct(self, **kwargs)
   1368 try:
-> 1369     obj = self.__new_container__(cls, builder.source, parent, builder.attributes.get(self.__spec.id_key()),
   1370                                  **kwargs)
   1371 except Exception as ex:

File ~/Documents/EEGLAB/EEG-Dash-Data/.venv/lib/python3.11/site-packages/hdmf/build/objectmapper.py:1382, in ObjectMapper.__new_container__(self, cls, container_source, parent, object_id, **kwargs)
   1380 # obj has been created and is in construction mode, indicating that the object is being constructed by
   1381 # the automatic construct process during read, rather than by the user
[...]

PR #652, opened 2025-01-30
Fix bug when exporting scalar text

ehennestad:
## Motivation
Fix bug discovered in issue #650

## How to test the behavior?
```
tests.unit.nwbExportTest/testExportScalarTextAttributeWithEmptyString
```

## Todo
- [x] Merge PR #655 

## Checklist


ehennestad:
> I'm not really a big fan of supporting empty strings at all. It seems the only time you would ever use an empty string would be to get around providing a required attribute, right?

I guess, but now those should be required (if a required property is empty, it would throw an error on export), so this PR might be obsolete. The default value for string property is an empty string, but I also think empty values are not exported.

In any case, this still fixes a bug in my opinion

bendichter (approved):

PR #655, opened 2025-02-01
Fix: Mark required attribute- and link-properties as required

### LINKED ISSUE #654: [Bug]: Required properties of NWB schemas are not marked as required in MatNWB
### What happened?

Properties originating from **Attributes** in the NWB specification schemas are not marked as required by MatNWB, even though they are required by the schemas.

This was discovered in issue #650 

@bendichter It is not obvious to me from the [NWB Specification Language](https://schema-language.readthedocs.io/en/latest/description.html#sec-link-spec), but are **Link** properties always required?

### Steps to Reproduce

```matlab
types.core.ProcessingModule() % does not issue a warning that description (required) is missing.
```

### Error Message

```matlab

```

### Operating System

macOS

### Matlab Version

2023b

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/neurodatawithoutborders/matnwb/blob/master/.github/CODE_OF_CONDUCT.md)

ehennestad:
Fix #654 

## Motivation

Attributes and links that were required according to the NWB schemas were not required in MatNWB. This PR fixes that.
It also adds a warning if a dependent required attribute is missing on export. For example, an attribute might be required but part of an untyped dataset or group which is not required. If such a dataset or group is added, but a required attribute of that dataset or group is not, a warning is shown on export.

## How to test the behavior?

```matlab
% Example 1 - Required attribute. Ok before, fails now

nwbFile1 = NwbFile( ...
    'session_description', 'test file for nwb export', ...
    'identifier', 'export_test', ...
    'session_start_time', datetime("now", 'TimeZone', 'local') );
processingModule = types.core.ProcessingModule();
nwbFile1.processing.set('ModuleWithoutDescription', processingModule);
nwbExport(nwbFile1, "example1.nwb")


% Example 2 - Required link. Fails before, but for the wrong reason

nwbFile2 = NwbFile( ...
    'session_description', 'test file for nwb export', ...
    'identifier', 'export_test', ...
    'session_start_time', datetime("now", 'TimeZone', 'local') );
electrode = types.core.IntracellularElectrode('description', 'test');
nwbFile2.general_intracellular_ephys.set('Electrode', electrode)
nwbExport(nwbFile2, "example2.nwb")

% Example 3 - Required dependent attribute. Ok before, issues warning now

nwbFile3 = NwbFile( ...
    'session_description', 'test file for nwb export', ...
    'identifier', 'export_test', ...
    'session_start_time', datetime("now", 'TimeZone', 'local') );
nwbFile3.general_source_script = '.../nwbExportTest.m';
nwbExport(nwbFile3, "example3.nwb")
```

## Checklist


bendichter:
I'm a bit confused by this one:

```matlab
% Example 2 - Required link. Fails before, but for the wrong reason

nwbFile2 = NwbFile( ...
    'session_description', 'test file for nwb export', ...
    'identifier', 'export_test', ...
    'session_start_time', datetime("now", 'TimeZone', 'local') );
electrode = types.core.IntracellularElectrode('description', 'test');
nwbFile2.general_intracellular_ephys.set('Electrode', electrode)
nwbExport(nwbFile2, "example2.nwb")
```

To be a link it must be added somewhere else in the NWB file, right?

bendichter:
```matlab
% Example 3 - Required dependent attribute. Ok before, issues warning now

nwbFile3 = NwbFile( ...
    'session_description', 'test file for nwb export', ...
    'identifier', 'export_test', ...
    'session_start_time', datetime("now", 'TimeZone', 'local') );
nwbFile3.general_source_script = '.../nwbExportTest.m';
nwbExport(nwbFile3, "example3.nwb")
```

what was the missing attribute?

+tests/+unit/nwbExportTest.m
```diff
@@ -38,6 +38,44 @@ function testExportNwbFileWithMissingRequiredProperties(testCase)
                 'NWB:RequiredPropertyMissing')
         end
 
+        function testExportNwbFileWithMissingRequiredAttribute(testCase)
+            processingModule = types.core.ProcessingModule();
```
bendichter, line 42:
```suggestion
            % this should raise an error because `ProcessingModule` requires a description
            processingModule = types.core.ProcessingModule();
```

bendichter:
And I guess these changes didn't cause any problems for the tutorials?

ehennestad:
> I'm a bit confused by this one:
> 
> ```matlab
> % Example 2 - Required link. Fails before, but for the wrong reason
> 
> nwbFile2 = NwbFile( ...
>     'session_description', 'test file for nwb export', ...
>     'identifier', 'export_test', ...
>     'session_start_time', datetime("now", 'TimeZone', 'local') );
> electrode = types.core.IntracellularElectrode('description', 'test');
> nwbFile2.general_intracellular_ephys.set('Electrode', electrode)
> nwbExport(nwbFile2, "example2.nwb")
> ```
> 
> To be a link it must be added somewhere else in the NWB file, right?

Yes, this is not very clear.
`types.core.IntracellularElectrode` has the property `device`, which is a Link. Before, this property was not required, and now it is. Running the above code should fail, because the linked `device` is not added.
If you test the example on the master branch, it will fail, but the error message is unspecific, and should be considered a bug. If you run it on the PR branch, the error message will state the the required property `device` is missing

ehennestad:
> ```matlab
> nwbFile3 = NwbFile( ...
>     'session_description', 'test file for nwb export', ...
>     'identifier', 'export_test', ...
>     'session_start_time', datetime("now", 'TimeZone', 'local') );
> nwbFile3.general_source_script = '.../nwbExportTest.m';
> nwbExport(nwbFile3, "example3.nwb")
> ```

`general_source_script_file_name `. Running example 3 should now show this warning

```
Warning: The property "general_source_script_file_name" of type "NwbFile" in file location "root" is required when the property
"general_source_script" is set. Please add a value for "general_source_script_file_name" and re-export. 
```

bendichter:
If `device` is required by `IntracellularElectrode`, then wouldn't it be better to flag that in the constructor of `IntracellularElectrode`? Is that possible? This will likely cause backward compatibility issues where users were able to create NWB files that were out of spec and want to read them now. We solved issues like this, where we became more strict about the form of data, by creating checks that are run when an object is being created in memory (which throw errors) and separate checks when a file is being read (which throw warnings). Do you think something like that might be possible here?

bendichter:
I guess the advantage of your approach is that if we flag these things on write then we don't have to worry about breaking read for old out-of-spec files. Since this is a pretty serious bug, I'm fine with pushing it forward in this way, even though it's not the idea solution. But ultimately it would be ideal to put these checks in the class constructors.

ehennestad:
@bendichter , The check will also occur on class construction. For example:

```matlab
>> types.core.IntracellularElectrode()

ans = 

  IntracellularElectrode with properties:

                  description: ''
                       device: []
                      cell_id: ''
                    filtering: ''
    initial_access_resistance: ''
                     location: ''
                   resistance: ''
                         seal: ''
                        slice: ''

Warning: The following required properties are missing for instance for type "types.core.IntracellularElectrode":
    description
    device 
```
Note that the warning does not appear if the output is suppressed, e.g.

```matlab
>> ielectrode = types.core.IntracellularElectrode();
>> 
```
and we could think about fine-tuning this, but I would suggest making that a separate issue. I would definitely go for warning on construction / error on export to be able to read files where these properties might be missing, but also as a more "gentle" way of teaching users what are required and not

bendichter:
The warning is good, but I don't like that it is suppressed by the semicolon. I think most users will use a semicolon here and so they will miss the warning. It is a very important warning, so I don't want them to miss it. I was not aware of this behavior in MATLAB. Is this custom or standard warning behavior?

bendichter (approved):

PR #657, opened 2025-02-01
Fix warn if property attribute not exported

ehennestad:
## Todo:
- [x] Merge PR #655 before this one

## Motivation

In MatNWB it is possible to set values for attribute properties of dataset properties where the dataset property is empty. If this is done, the attribute value is not exported to file.

In PR #596, we added a warning for such cases. However some of the attributes in question have default values, and the warning was shown for those attributes as well, even though the user has not entered a value for the property.

This PR fixes that so that a warning is only displayed if the user has entered a value which is different than default for an attribute property which is not exported to file

## How to test the behavior?
```matlab
nwbFile = NwbFile( ...
    'session_description', 'test file for nwb export', ...
    'identifier', 'export_test', ...
    'session_start_time', datetime("now", 'TimeZone', 'local') );

% Add device to nwb object
device = types.core.Device();
nwbFile.general_devices.set('Device', device);

imaging_plane = types.core.ImagingPlane( ...
    'device', types.untyped.SoftLink(device), ...
    'excitation_lambda', 600., ...
    'indicator', 'GFP', ...
    'location', 'my favorite brain location');
nwbFile.general_optophysiology.set('ImagingPlane', imaging_plane);

% Before: Issues warning
% After: No warning (Fixed)
nwbExport(nwbFile, 'test_1.nwb')

% Change value for attribute of the grid_spacing dataset.
% Because grid_spacing is not set, this attribute value is not
% exported to the file.
imaging_plane.grid_spacing_unit = "microns";

% Before: Issues warning
% After: Issues warning
nwbExport(nwbFile, 'test_2.nwb')
```

## Checklist


bendichter:
Could you throw a warning on this line 

`imaging_plane.grid_spacing_unit = "microns";`

instead of write?

ehennestad:
> Could you throw a warning on this line
> 
> `imaging_plane.grid_spacing_unit = "microns";`
> 
> instead of write?

Update. There will now be a warning both on property assignment as you suggest and again on export if the condition is still true

PR #670, opened 2025-02-16
Fix: Cache with map of required properties per neurodata type must be version specific

ehennestad:
## Motivation

Required properties of neurodata types might change across NWB schema versions. This PR fixes a bug where an internal cache storing required properties for neurodata types did not distinguish between different schema versions

## How to test the behavior?
Code snippet that fails before, but works after.
 
```matlab
generateCore('2.0.2')
NwbFile()
generateCore();
NwbFile()
```

## Checklist


bendichter (approved):

+types/+untyped/MetaClass.m
ehennestad, line 1: First three changes in this file is just breaking up very long lines

PR #675, opened 2025-03-03
Support adding string type as value to neurodata object properties 

ehennestad:
## Motivation

Adding string types to properties causes error on nwbExport. As the string type was introduced in R2016b, it should be supported by MatNWB.

## How to test the behavior?
```matlab
nwb = tests.factory.NWBFile();

generalExperimenter = ["John Doe", "Jane Doe"];
generalExperimentDescription = "Test with string data types";
nwb.general_experimenter = generalExperimenter;
nwb.general_experiment_description = generalExperimentDescription;

nwbExport(nwb, "test.nwb");
```

## Checklist


ehennestad:
NB:
2 tests are failing because the Clustering neurodata type was deprecated in pynwb, but is still used in the matnwb test suite

PR #681, opened 2025-03-07
Add functions for inferring directory of generated types

ehennestad:
~~**Note**: Tests fails because of update to nwb inspector. Fixed in PR #683~~ Merged
## Motivation
Generated neurodata type classes and format specifications (namespace) caches are saved in the root MatNWB directory by default.

However, it is possible to generate types and caches in other locations by specifying a `savedir` argument when running `generateCore`/`generateExtension`  (for example temporary locations which can be useful during testing).

For the function `schemes.loadNamespace`, the default behavior was to look for the format specifications (namespace) caches in the root MatNWB directory. This PR changes this default behavior to instead find the directory where the `types.core.NWBFile` is located by looking at MATLAB's search path.

This makes the `schemes.loadNamespace` function more robust when used during testing where the namespace caches might be present in a different location than the root MatNWB directory.

## Checklist


+schemes/+utility/findRootDirectoryForGeneratedTypes.m
```diff
@@ -0,0 +1,33 @@
...
+        ME = MException('NWB:Types:GeneratedTypesNotFound', ...
+            ['Could not find a location for generated classes of neurodata ', ...
+            'types. Please check if MatNWB is properly added to MATLAB''s ' ...
+            'search path and/or run generateCore() to re-generate classes.'] );
+        throwAsCaller(ME)
+    end
+    
+    if numel(generatedTypeLocation) > 1
+        warning('NWB:Types:MultipleGeneratedTypesFound', ...
+            ['Multiple generated types was found for %s.\n', ...
```
bendichter, line 24:
```suggestion
            ['Multiple generated types were found for %s.\n', ...
```

bendichter: Could you indicate the paths for each?

ehennestad: Added display of paths in the warning message

bendichter (approved):

PR #688, opened 2025-03-20
Fix listing of required properties for neurodata types.

ehennestad:
## Note
Depends on PRs
- #695 
- #696 

## Motivation
This PR fixes bugs related to required properties of neurodata types:
- List of required properties in documentation (docstrings/read the docs) does not match actual required properties of neurodata types.
- Some required properties of neurodata types were not correctly resolved from NWB schemas
- The current way of resolving required properties was not compatible with MATLAB releases earlier than R2022a

This PR creates a single function for resolving required properties, and uses it both for adding the required properties to the neruodata type class' docstring and for checking if required properties are missing on object construction / nwbExport

## Checklist


ehennestad:
I have looked at a few schemas, and this looks accurate. However, would like to know if there is an efficient way to check all schemas

bendichter:
Are there specific data requirements that weren't being enforced before? Could you add one of those as a new test?

ehennestad:
Added a test that would have failed before

bendichter (approved):

PR #695, opened 2025-03-25
Fix parsing of quantity specification value

ehennestad:
## Motivation

The NWB specification language [supports multiple value options](https://schema-language.readthedocs.io/en/latest/description.html#quantity) for specifying the quantity of Groups, Datasets and Links, but currently MatNWB will only parse the following quantity values: `*`, `+`, `?`. In other words, it does not support specifying exact quantities as numeric values, or the full text options (`zero_or_many`, `one_or_many`, `zero_or_one`)

This PR adds cases for parsing all possible allowed values as well as error handling if the quantity specifier is invalid.

Changes in this PR:
- Add a class `file.interface.HasQuantity` that can be mixed in with specification types that has a `quantity` specifier
- Add `file.interface.HasQuantity` as superclass to `Group`, `Dataset` and `Link` specification types
- Add a test specification schema and a unit test for testing that quantity specification values are correctly determining if a datatype property should be required or not.

## Checklist

### DIFF
--- a/+file/+interface/HasQuantity.m
+++ b/+file/+interface/HasQuantity.m
@@ -1,0 +1,84 @@
+classdef HasQuantity
+% HasQuantity - Provide methods for parsing quantity specification value
+    
+    % properties
+    %     QuantityKey = 'quantity'
+    % end
+    
+    methods (Static, Access = protected)
+        function isRequired = isRequired(source)
+            if isKey(source, 'quantity')
+                quantity = source('quantity');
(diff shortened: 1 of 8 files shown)


bendichter:
Can you confirm that if a container specifies a quantity of a specific number (e.g. 2) then that number is enforced?

bendichter (approved):

bendichter:
ok so this is specifically about reading HDMF. Looks fine then

PR #696, opened 2025-03-27
Add test specification and test suite for testing required props

ehennestad:
## Motivation

Add a test specification and a test suite for testing proper internal resolution of required properties.

This added test suite should check that required properties are correctly resolved for groups, datasets, links and attributes. It also tests that required properties are correctly resolved if the "required" status of a property is overridden in a data type that extends another type.

**Note: Some of these tests will fail, but failures are fixed by PR #688** 

## How to test the behavior?
See test results

## Checklist


bendichter:
looks great! Let's merge 688 and make sure all of this passes

ehennestad:
Closed as these changes were already merged as part of PR #688

PR #702, opened 2025-04-03
Add DandiHub edition of read NWB file tutorial

vijayiyer05:
Minimally modified `read_demo` tutorial that 1) uses `dandi `Python package for download and 2) bypasses MatNWB install.

## Motivation

The original tutorial was slow getting to the science, with initial sections focused on downloading the data and installing MatNWB. In the DandiHub configured environment, the former can be streamlined (by using the pre-installed `dandi` Python package) and the latter can be skipped (since MatNWB is pre-instlalled). 

## How to test the behavior?
```

1. Select the MATLAB image for the Base server option on DandiHub. 
2. Supply MathWorks credentials if prompted (not needed every time)
3. When the launcher appears, select the MATLAB desktop option
4. Open this live script (in the MatNWB tutorials folder) and select Run (▶️button) from the Live Editor tab in the desktop ribbon

The live script should run through and generate the same outpus as in the saved live script. The downloaded data should appear in the dandisets\000004 folder and is ~70MB in size.
```

## Checklist

- [x] Have you ensured the PR description clearly describes the problem and solutions?
- [x] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/neurodatawithoutborders/matnwb/pulls) for the same change?
- [x] If this PR fixes an issue, is the first line of the PR description `fix #XX` where `XX` is the issue number?

PR #704, opened 2025-04-25
Add DandiHub edition of read NWB file tutorial (modified)

ehennestad:
## Motivation

See PR #702 

This PR contains the following modifications:

- Adds an "environment" selector making it possible to run the py.dandi.download also in a local environment
- Removes a comment in the nwbRead section about the ignorecache flag as this is not relevant here
- Adds .html and .m versions of livescripts
- Ignore the new livescript when testing tutorial live scripts
- Add the new tutorial to the ReadTheDocs page

Preview livescript here:
https://matnwb--704.org.readthedocs.build/en/704/pages/tutorials/read_demo_dandihub.html

**Note**: The environment variable is a dropdown selector in the livescript. The default value for the environment is DandiHub, but for creating a html locally, I ran it with the value set to Local.


vijayiyer05:
@ehennestad, I like this approach with a Live Control, and I think it makes sense to proceed adding this as a v2.8 patch. 

FYI  for the future: there's been a first attempt to make a standard recipe of this type as part of the NWB/DANDI [example library](https://github.com/MATLAB-Community-Toolboxes-at-INCF/example-live-scripts/tree/main), with at least a best default based on detected environment. We might consider how to update those, and where if anywhere to centrally collect such recipes, down the road. Note future DANDIHub updates (wrt MATLAB environment) are currently envisioned at each MATLAB release timepoint, i.e., next chance in the month or so folloiwng R2025a, then again R2025b, etc.

bendichter (approved):

PR #705, opened 2025-05-01
Add mixin class for data types with un-named (sub)group properties, i.e ProcessingModule

### LINKED ISSUE #625: [Feature]: Change implementation of containers types to simplify API

Many NWB types are containers for other types, with two examples being `ProcessingModule` and `Fluorescence`.

The syntax for setting/getting data for these containers is cumbersome:
```
ophys_module = types.core.ProcessingModule('description', 'contains optical physiology data')

ophys_module.nwbdatainterface.set('ImageSegmentation', img_seg);
```
and 

```
fluorescence = types.core.Fluorescence();

fluorescence.roiresponseseries.set('RoiResponseSeries', roi_response_series);
ophys_module.nwbdatainterface.set('Fluorescence', fluorescence);
```

A simpler way of doing this would be:
```
ophys_module = types.core.ProcessingModule('description', 'contains optical physiology data')

ophys_module('ImageSegmentation') = img_seg;
```
and
```
fluorescence = types.core.Fluorescence();

fluorescence('RoiResponseSeries') = roi_response_series;
ophys_module('Fluorescence') = fluorescence;
```

A similar suggestion was raised in pynwb ([issue #1993](https://github.com/NeurodataWithoutBorders/pynwb/issues/1993)) and it would be great to mirror this on the matnwb side.

_No response_

Redefine container classes to have `dictionary`/`Set` like behavior, while also restricting the types it is allowed to hold

Yes.

### Code of Conduct

ehennestad:
## Motivation
Fix #625 

## How to test the behavior?
```
```

#### Adding data types to a container type

- Data type is added to a container, i.e `aProcessingModule.add('namedDataType')`
  - Is the data type added to the underlying types.untyped.Set?
- Is the data type added as a dynamic property to the container type (i.e a ProcessingModule) on `nwbRead`?

#### Removing data types from a container type
todo

## Checklist

- [ ] Have you ensured the PR description clearly describes the problem and solutions?
- [ ] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/neurodatawithoutborders/matnwb/pulls) for the same change?

### DIFF
--- a/+file/fillClass.m
+++ b/+file/fillClass.m
@@ -92,6 +92,12 @@ function template = fillClass(name, namespace, processed, classprops, inherited,
         classTag = 'types.untyped.DatasetClass';
     end
 
+    if isa(class, 'file.Group')
+        if class.hasAnonGroups
+            classTag = sprintf('%s & matnwb.mixin.HasGroups', classTag);
+        end
+    end
+
     %% return classfile string
     classDefinitionHeader = [...
         'classdef ' name ' < ' depnm ' & ' classTag newline... %header, dependencies
@@ -128,6 +134,7 @@ function template = fillClass(name, namespace, processed, classprops, inherited,
         , '% OPTIONAL PROPERTIES' ...
         } ...
         );
+
     fullPropertyDefinition = '';
     for iGroup = 1:length(PropertyGroups)
         Group = PropertyGroups(iGroup);
@@ -140,6 +147,17 @@ function template = fillClass(name, namespace, processed, classprops, inherited,
             , Group.Separator ...
             , propertyDefinitionBody ...
             }, newline);
+    end
+    if isa(class, 'file.Group')
+        if class.hasAnonGroups
+            isAnonGroup = arrayfun(@(x) isempty(x.name), class.subgroups, 'uni', true);
+            anonNames = arrayfun(@(x) lower(x.type), class.subgroups(isAnonGroup), 'uni', false);
+            fullPropertyDefinition = strjoin({...
+                fullPropertyDefinition, ...
+                '    properties (Access = protected)', ...
+                sprintf('        GroupPropertyNames = {%s}', strjoin(strcat('''', anonNames, ''''), ', ') ), ...
+                '    end'}, newline);
+        end
     end
 
     constructorBody = file.fillConstructor(...
--- a/+matnwb/+mixin/HasGroups.m
+++ b/+matnwb/+mixin/HasGroups.m
@@ -1,4 +1,4 @@
-classdef HasGroups < matlab.mixin.CustomDisplay & matlab.mixin.indexing.RedefinesParen & handle
+classdef HasGroups < matlab.mixin.CustomDisplay & matlab.mixin.indexing.RedefinesDot & handle
 % HasGroups - Provides methods for retrieving group elements by their key names 
 %
 % This mixin class allows accessing elements in Set properties using parentheses notation.
@@ -12,14 +12,14 @@ classdef HasGroups < matlab.mixin.CustomDisplay & matlab.mixin.indexing.Redefine
     end
     
     methods (Access = protected)
-        function varargout = parenReference(obj, indexOp)
+        function varargout = dotReference(obj, indexOp)
             % Handle parentheses indexing references
             % Check if the index is a string (key name)
 
-            key = indexOp(1).Indices{1};
+            key = indexOp(1).Name;
 
-            if length(indexOp) == 1 && ischar(key)
-                keyName = indexOp(1).Indices{1};
+            if ischar(key) || isstring(key)
+                keyName = indexOp(1).Name;
                 
                 % Check if the key name matches a key in any of the Set properties
                 for i = 1:length(obj.GroupPropertyNames)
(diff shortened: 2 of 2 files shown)


ehennestad:
This is currently possible in MatNWB, but creates an invalid group in the NWB-file where two different data type objects are merged in the same ground:

```matlab
nwbFile = tests.factory.NWBFile();
module = types.core.ProcessingModule();

nwbFile.processing.set('test', module);

T = struct2table(struct('name', {'a', 'b'}, 'value', {1, 2}));
dynamicTable = util.table2nwb(T);

nwbFile.processing.get('test').description = 'a test';
nwbFile.processing.get('test').nwbdatainterface.set('test', tests.factory.TimeSeriesWithTimestamps)
nwbFile.processing.get('test').dynamictable.set('test', dynamicTable)

nwbExport( nwbFile, 'test_same_name_in_different_subgroups.nwb')
```

Easy to prevent for this new syntax, but not for the legacy syntax.

+file/fillClass.m
ehennestad, line 1: - renamed `depnm` to `superclassNames`
- add `matnwb.mixin.HasUnnamedGroups` as superclass to data types with un-named subgroups

+schemes/+utility/listGeneratedTypes.m
ehennestad, line 1: Convenience method for listing class names (or short names) of all generated types

bendichter:
Do we need all of this renaming logic? What if someone tries to add two objects, `Time-Series` and `Time_Series`?

Why can't we just change `module.datainterfaces.get("Foo")` to `module.get("Foo")` or `module("Foo")`? Then we don't need to mess with any of this name changing stuff. We can expose `module.Foo` if the name allows it, and if not then revert to the other syntax.

ehennestad:
> Do we need all of this renaming logic? What if someone tries to add two objects, Time-Series and Time_Series?

That will look like this:
```matlab
module = types.core.ProcessingModule('description', 'a module with two timeseries');
module.add('time_series', types.core.TimeSeries)
module.add('time-series', types.core.TimeSeries)

>> module

module = 

Warning: The following named elements of "ProcessingModule" are remapped to have valid MATLAB names, but will be written to file with
their actual names:
       ValidName        ActualName
    _______________    _____________

    "time_series_1"    "time-series"
 
  ProcessingModule with properties:

      description: 'a module with two timeseries'
    time_series_1: [1×1 types.core.TimeSeries]
      time_series: [1×1 types.core.TimeSeries]
```
      
      
Which is actually quite similar to how MATLAB does it in e.g readtable.

given a test.csv file:
```csv
time-series	time_series
1	2
3	4
```

```matlab
>> readtable('test.csv')
Warning: Column headers from the file were modified to make them valid MATLAB identifiers before creating variable names for the
table. The original column headers are saved in the VariableDescriptions property.
Set 'VariableNamingRule' to 'preserve' to use the original column headers as table variable names. 

ans =

  2×2 table

    time_series    time_series_1
    ___________    _____________

         1               2      
         3               4      
```


> Why can't we just change module.datainterfaces.get("Foo") to module.get("Foo") or module("Foo")? Then we don't need to mess with any of this name changing stuff. We can expose module.Foo if the name allows it, and if not then revert to the other syntax.

First, I would like to avoid using `module.get` because its not a very common MATLAB syntax, and it's difficult for a user to know when to use `module.get` and when to use standard dot-indexing, i.e `module.desription`.

Implementing parenthesis indexing is quite straightforward with the `matlab.mixin.indexing.RedefinesParen` class, but it is only supported from R2021b. It is possible to override `subsref` for compatibility in older releases, but I would really not recommend doing that.

Also, one syntax for "valid" names and another syntax for other names I think would be confusing.


In summary, the renaming logic would provide full support for reading files from pynwb where names are specified with spaces or or other symbols that are not supported in MATLAB but also take care of edge cases like the one you use as an example.

bendichter:
You're not warning on the add command? I'm really not into magically renaming. Now the order in which the data elements are added to the file object matters and the user needs to learn the renaming rule. I think `module.get('my-name')` is so much simpler and sufficiently solves the problem without getting into tricky territory. I understand it's a bit unusual syntax but I've just been in too many situations where renaming causes more confusion than it is worth.

ehennestad:
> You're not warning on the add command?

Good point, there could be an extra warning on the add command.

> I'm really not into magically renaming

It's not really renaming, just creating an alias which is used for the dynamic properties that enables dot-indexing and autocompletion. `module.nwbdatainterface.get('my-name')` would still work and creating a shortcut as `module.get('my-name')` would be a quick addition.

How about keeping both modes?

A typical MATLAB user would likely specify names using valid camelCase or snake_case and not encounter the name remapping at all. If they loaded a file created with python using different naming convention, their preferred modes might vary. For example, I prefer to explore new objects in the command window, in which the alias-names are presented and where I can use them for dot-indexing. If someone else prefers to use the "real" names they could do so with `module.get('my-name')`

bendichter:
This is the old syntax: `nwb.processing.get('ophys').nwbdatainterface.get('Fluorescence').roiresponseseries.get('RoiResponseSeries')` but what if you had an RoiResponseSeries named "roiresponseseries"? I think the only solution is in this case to require the syntax `nwb.processing.get('ophys').nwbdatainterface.get('Fluorescence').roiresponseseries.get('roiresponseseries')`

+tests/+unit/HasUnnamedGroupsTest.m
```diff
@@ -0,0 +1,276 @@
...
+                    'TimeSeries', types.hdmf_common.DynamicTable());
+            
+            testCase.verifyError(operationToVerify, ...
+                'NWB:HasUnnamedGroups:DuplicateEntry')
+        end
+        
+        function testOverrideExistingEntry(testCase)
+            module = types.core.ProcessingModule(...
+                'description', 'test module', ...
+                'Test', types.core.TimeSeries());
```
bendichter: what happens if you have a TimeSeries named "description"?

docs/source/pages/how_to/working_with_containers.rst
```diff
@@ -0,0 +1,433 @@
...
+
+Handling invalid MATLAB names
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+MATLAB property names must follow specific rules (no spaces, special characters, must start with a letter, etc.). When you add a data object with a name that isn't a valid MATLAB identifier, MatNWB automatically creates a valid alias:
+
+.. code-block:: MATLAB
+
+    % Add with spaces and special characters
+    processingModule.add('my time series', timeSeries);
```
bendichter: what happens if you add two objects with names that resolve to the same alias?

ehennestad: Covered in this section:
https://github.com/NeurodataWithoutBorders/matnwb/blob/d49bc3c0e3818e4dccf4b6360c117bb172ab4b0f/docs/source/pages/how_to/working_with_containers.rst?plain=1#L235-L251

+tests/+unit/HasUnnamedGroupsTest.m
ehennestad: That causes an error:

``` matlab
Error using assert
Property with name "description" already exists
```

PR #714, opened 2025-05-08
Update fillValidators.m

### LINKED ISSUE #712: [Bug]: Validation of data for subclasses of PatchClampSeries are missing.
### What happened?

Example where the function `validate_data` of `types.core.CurrentClampSeries` is empty.

https://github.com/NeurodataWithoutBorders/matnwb/blob/16862b7958058fe917f1a06b6a7d67d2de1b4b0f/%2Btypes/%2Bcore/CurrentClampSeries.m#L156-L158

This appears to be the case for the following classes:
- `types.core.CurrentClampSeries`
- `types.core.CurrentClampStimulusSeries`
- `types.core.VoltageClampSeries`
- `types.core.VoltageClampStimulusSeries`

### Steps to Reproduce

```matlab
N/A
```

### Error Message

```matlab

```

### Operating System

macOS

### Matlab Version

Any

### Code of Conduct

ehennestad:
Fix #712
Fix issue where dimension validation was dependent on datatype

## Motivation

Fixed a bug where the class generator code would skip filling in `shape` validation of a dataset if the `dtype` is "any".

This change also necessitated an update to the data type specification processing, where inherited specification fields needed to be resolved. MatNWB would previously use the default value for a specification key instead of an inherited value if the value was specified on the parent/ancestor type but not on the current type. 

## How to test the behavior?
```
See updated type classes
```

## Todo

- [ ] Add a unittest
- [x] Merge [`712-missing-data-validators`](https://github.com/NeurodataWithoutBorders/matnwb/tree/712-missing-data-validators) into this PR 
## Checklist


ehennestad:
This will affect the `data` property of the following types, where the `dtype` is any:

- CurrentClampSeries
- CurrentClampStimulusSeries
- TimeSeries
- VoltageClampSeries
- VoltageClampStimulusSeries

Before the fix, these classes can have any shape. For example, a TimeSeries can have 5 or higher dimensional data, even though the schema specifies it can have 1-4D data. 

However, after the fix, the following classes will be restricted to a scalar shape, because of how inheritance of HDMF properties was defined in the past. See this issue for more details: https://github.com/NeurodataWithoutBorders/nwb-schema/issues/622#issuecomment-2863911940 
 
- CurrentClampSeries
- CurrentClampStimulusSeries
- VoltageClampSeries
- VoltageClampStimulusSeries

I looked into changing how inheritance rules are applied in this commit:
https://github.com/NeurodataWithoutBorders/matnwb/compare/main...712-missing-data-validators

Need to merge this into this branch, but should we wait for the inheritance rules to be official on HDMF (according to this issue: https://github.com/hdmf-dev/hdmf/issues/320) ?

+schemes/+internal/findMatchingParentSpec.m
```diff
@@ -0,0 +1,39 @@
+function matchingParentSpec = findMatchingParentSpec(spec, parentSpecs)
+% findMatchingParentSpec - Find a spec from a parent specification set
+% matching on name or "neurodata_type_inc"
+
+    % The order here is important. All base types (i.e., groups, datasets, 
+    % attributes, and links) can have a name. However, for groups and datasets,
+    % the name is optional — if omitted, the specification identifier will 
+    % default to the `neurodata_type_inc` value instead.
+    ID_KEYS = ["name", "neurodata_type_inc"];
```
ehennestad, line 9: I needs some help here @bendichter @rly @oruebel 

This function will be used in MatNWB to find a matching specification (i.e for an attribute, dataset, group or link) from the list of attributes, datasets, groups or links defined in a parent datatype.

I initially thought to match based on the `name` key, and if there is no name (i.e for datasets, groups) to fall back to the `neurodata_type_inc` key. But I am wondering if that fallback makes sense? I.e is there any scenarion where a non-named base type can be redefined in a sub data type?

bendichter (approved):

PR #720, opened 2025-06-12
Update test workflow to run lighter testing on draft PRs

ehennestad:
## Motivation

Speed up testing when a PR is in draft mode

## How to test the behavior?
```
N/A
```

## Checklist

- [ ] Have you ensured the PR description clearly describes the problem and solutions?
- [ ] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/neurodatawithoutborders/matnwb/pulls) for the same change?


ehennestad:
A few changes to speed things up:

1. only run tests with latest matlab release (no matrix) if PR is in draft mode

2. cache matlab setup and python dependencies 

no tests are skipped

sometimes testing on all releases can take more time than usual (ie if some workers are queues or for other reasons take longer to finish) last week a test run took 27 min compared to usual 12.

Caching saves about 1 minute

bendichter (dismissed):

ehennestad:
Based on discussion:
- Added more comments in the set-matrix workflow job
- Removed the part adding EOF before and after a json string
- Removed caching for pip dependencies. No noticable speedup + risk of using oudated versiond of dependencies

PR #730, opened 2025-07-07
Documentation: Add copy button to code blocks

ehennestad:
## Motivation

This PR adds several improvements to the tutorial livescript HTMLs that are embedded in the online documentation:

- Copy button for code blocks
- Reactive CSS adjusting width of embedded images to width of page
- Ensure all code blocks are the same width
- Fixed clipping of text in some output containers of code blocks
- Fixed relative links pointing from one tutorial to another. Tutorial html files are embedded in iframes, and the old behavior would open the link within the same iframe, now it will open actual tutorial page in the docs.

**Note:** The following tutorials are not updated:
- Basic Usage of MatNWB (basicUsage.mlx)
- Converting Trials to NWB Format (convertTrials.m)

## How to test the behavior?
```
Check documentation built in checks of this PR
```

## Checklist

### DIFF
--- a/+file/Attribute.m
+++ b/+file/Attribute.m
@@ -11,6 +11,7 @@ classdef Attribute
         scalar; %if the value is scalar or an array
         dimnames;
         shape;
+        isConstrainedSet;
     end
     
     methods
@@ -27,6 +28,8 @@ classdef Attribute
             obj.scalar = true;
             obj.shape = {};
             obj.dimnames = {};
+            obj.isConstrainedSet = false; % Always false for attributes
+
             
             if nargin < 1
                 return;
--- a/+file/Link.m
+++ b/+file/Link.m
@@ -4,6 +4,8 @@ classdef Link < file.interface.HasQuantity
         name;
         required;
         type;
+        scalar;
+        isConstrainedSet
     end
     
     methods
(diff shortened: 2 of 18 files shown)


bendichter:
Is there any way to avoid this?

image


\

ehennestad:
Some of the tutorials already have overflowing text in the current version of the docs, and I was planning to re-format the code in separate PRs to avoid very long lines. The current `icephys` tutorial does not have overflowing text, but the code block container wraps the content and some code blocks are larger than the width of the ReadTheDocs page. 

Formatting code would be quick, and it would at look ok on most computer screens (although not mobile or if resizing the browser to be very narrow).

**There are a few other options:** 
1. Use the css rules of the `icephys` tutorial in the current docs. This is quick, but I think it looks weird when code blocks are different sizes, and potentially requires horizontal scrolling to see very long lines and copy buttons
2. Allow scrolling inside the code boxes. However, the code block outline/border is part of nested containers of the exported html, and would move with the scroll. This looks weird. Not really an option, just mentioning it as a prelude to next option.
3. Postprocess livescript HTMLs to improve the organization of nested containers to allow an outer container with a border, and scrollable containers inside. This would be the best solution, but postprocessing autogenerated htmls is most likely take  a lot of time, and will break as soon as MathWorks changes the way they export htmls.

My preference is to reformat the code, or option 1 above.

ehennestad:
Facing two more issues:

Copying codeblock carries over some whitespace characters that are not valid Matlab characters
Copying codeblock with output in the next block will also copy the first line of that output.

PR #742, opened 2025-08-08
Fix dataset.data validation

### LINKED ISSUE #741: [Bug]: `DynamicTableRegion` accepts multidimensional `data`, violating NWB schema
### What happened?

According to the HDMF schema, the `DynamicTableRegion` type is defined as:
https://github.com/hdmf-dev/hdmf-common-schema/blob/31215f45b5106fddcf5042a9a4ed5ae9211e8936/common/table.yaml#L71-L83

```yaml
- data_type_def: DynamicTableRegion
  data_type_inc: VectorData
  dtype: int
  doc: DynamicTableRegion provides a link from one table to an index or region of another. The `table` attribute is a
    link to another `DynamicTable`, indicating which table is referenced, and the data is int(s) indicating the row(s)
    (0-indexed) of the target array. `DynamicTableRegion`s can be used to associate rows with repeated meta-data without
    data duplication. They can also be used to create hierarchical relationships between multiple `DynamicTable`s.
    `DynamicTableRegion` objects may be paired with a `VectorIndex` object to create ragged references, so a single cell
    of a `DynamicTable` can reference many rows of another `DynamicTable`.
  dims:
  - num_rows
  shape:
  - null
```

This implies that the `data` field should be **1-dimensional**, i.e., a vector of integers.

However, the following MATLAB code is currently accepted without error:

```matlab
dtr = types.hdmf_common.DynamicTableRegion('data', reshape(1:25, 5, 5));
```

This creates a `DynamicTableRegion` with a **5×5 matrix**, which violates the `shape: [null]` constraint defined in the schema.

### Error Message

```matlab
No error
```

### Expected Behavior

MatNWB should reject this input or raise a validation error indicating that the `data` field must be a 1-dimensional array.

This bug is present for all neurodata types that 1) are derived from NWBData and 2) have a shape constraint.

### Operating System

macOS

### Matlab Version

R2024b

### Code of Conduct

ehennestad:
Fix #741

Depends on PR #743 

## Motivation

Validation of dataset shape was not filled out by the class generation for the `data` property of neurodata types defined as dataset classes. This PR fixes that by ensuring shape information is added to the internal structure that represents information about a dataset property.

## Changes
- **Fixes shape validation of the `data` property for the following data types:**
  - `types.core.BaseImage`  
  - `types.core.ExternalImage`  
  - `types.core.GrayscaleImage`  
  - `types.core.Image`  
  - `types.core.ImageReferences`  
  - `types.core.NWBData`  
  - `types.core.RGBAImage`  
  - `types.core.RGBImage`  
  - `types.core.ScratchData`  
  - `types.core.TimeSeriesReferenceVectorData`  
  - `types.hdmf_common.Data`  
  - `types.hdmf_common.DynamicTableRegion`  
  - `types.hdmf_common.ElementIdentifiers`  
  - `types.hdmf_common.VectorData`  
  - `types.hdmf_common.VectorIndex`  
  - `types.hdmf_experimental.EnumData`  

- Support `shape: null` in specifications to mean any shape is allowed.
  - Fixed test schemas for multipleConstrained and regionReference (tests started failing based on this PR change) 

- Added function `applySchemaVersionPatches` to support current behavior of ScratchData (and VectorData/VectorIndex for earlier NWB versions) (NWB specification has not defined a shape, but it should have)

## How to test the behavior?
```
```

## Checklist


ehennestad:
This PR causes some tests to fail, due to inconsistencies in the schema specifications and the intended used of data types.
1. Tests of tutorials using `ScratchData`. `ScratchData` does not have a specified shape according to the schema, and defaults to a scalar shape (**2 tests failing**)
2. `VectorData` for v2.0.2 - v2.2.1  does not have a specified shape according to the schema, and defaults to a scalar shape (**4 tests failing**).

ehennestad:
> This PR causes some tests to fail, due to inconsistencies in the schema specifications and the intended used of data types.
> 
> 1. Tests of tutorials using `ScratchData`. `ScratchData` does not have a specified shape according to the schema, and defaults to a scalar shape (**2 tests failing**)
> 2. `VectorData` for v2.0.2 - v2.2.1  does not have a specified shape according to the schema, and defaults to a scalar shape (**4 tests failing**).

Added a function `applySchemaVersionPatches` to fix these cases by setting `shape` to `any` before generating the classes.

PR #794, opened 2026-03-24
Support nwb-schema v2.10.0

ehennestad:
## Motivation

Add support for nwb-schema v2.10.0.

### Changes in this PR
- Include all schemas for nwb-schema `v2.10.0` and hdmf-common-schema `v1.9.0`
- Regenerate class types
- Regenerate rst for neurodata / hdmf types
- Added mechanism for remapping the name given by schema before adding it as a class property. This was necessary for the new `events` field in `NWBFile` because `events` is a reserved keyword in MATLAB which can not be used for property names. Using remapping `events` -> `events_`

## How to test the behavior?
```
```

## Todo
- [x] ~~Pending release of NWB schema 2.10.0~~
- [x] ~~Add `addMeanings` utility method in the `types.util.dynamictable` namespace and write it into the `DynamicTable` class on type generation. Deferred to issue #846~~
- [x] `HERD` moves into `hdmf_common` namespace. PR #815 fixing a runtime validation bug in the `HERD` class should be merged.
- [x] Rerun `matnwb_generateRstFilesFromCode` to ensure new types and moved types are correctly represented in the docs

## Checklist


ehennestad:
The new `meanings_tables` property of the `DynamicTable` class is functionally equivalent to e.g the `acquisition` group on the NWB file. 

However, some extra validation should occur when adding a MeaningsTable.
1. The name of the entry should have the postfix "_meanings" and the suffix should match one of the named elements in the `colnames` property.
2. The linked `VectorData` object in the `target` property must match the `VectorData` object in the DynamicTable with the name matching the suffix of the `MeaningsTable` object's name.
3. All unique values of the linked `VectorData` must be present in the `VectorData` object stored in the `values` property. This validation might belong in the `MeaningsTable` constructor or the `validate_values` method of the `MeaningsTable` class.

To add `MeaningsTables` to a `DynamicTable` we will add a convenience method called `addMeanings` to ensure these extra constraints are checked. However, if someone  calls the `add` or `set` method directly on the `meanings_tables` **Set** on the `DynamicTable` these extra checks are not present. The validations above should also be added as a custom hook in the `validate_meanings_tables` method of the `DynamicTable`.

ehennestad:
### Reading `HERD` (v2.10.0) fails: base `Data` scalar-shape validation

**Reference issue:** https://github.com/hdmf-dev/hdmf-common-schema/issues/83

**Symptom:** `nwbRead` of a v2.10.0 file containing a `HERD` (`/general/external_resources`) errors — `types.hdmf_common.Data.validate_data` enforces shape `{[1]}` (scalar), but `entities` is a compound `[N×1]` table.

**Cause:** MatNWB defaults a dataset with no `shape` to scalar (`+file/Dataset.m`, `obj.shape = 1`) according to the hdmf-schema-language v2.0.2. Base `Data` (abstract; no dtype/shape/dims) therefore validates as scalar. HERD refines the included `Data` (`entities`, `keys`, …) to 1-D tables, but on read MatNWB reuses the base `Data` class, whose scalar check fires at construction — before HERD's own (correct) refined validation in `validate_entities` ever runs.

**Spec vs. hdmf-implementation:**
- hdmf-schema-language 2.0.2: a dataset with no `shape` defaults to **scalar**.
- HDMF's implementation does **not** hard-enforce this:
  - Validator is permissive — `check_shape(None, received) → True` for any shape (`src/hdmf/validate/validator.py`).
  - "Scalar expected" is applied **softly on read**: a 1-element **non-compound** dataset is unwrapped to a scalar (`src/hdmf/build/objectmapper.py`); compound datasets (like `entities`) are explicitly excluded.
  - For types added by inclusion, they resolve to the refined shape (`_resolve_inc_spec_shape`), so `entities` is validated as 1-D — never against base `Data`'s no-shape.
  
  
I will add an ad-hoc bypass for the MatNWB class generator, if type is hdmf_common.Data, allow any shape. This is similar to what HDMF is already doing. Does that sound ok, @rly, @bendichter ?

See this commit: a8b1ba68aef39a9857d08d907b6b691e13533760

ehennestad:
Note: @bendichter 

The generated `HERD` class can be used to create `HERD` objects, but it is currently not usable in practice, because MatNWB blacklists the `object_id` attribute on reads. So basically, neurodata type objects have no persistent uuid, and the mechanism HERD uses for tracking object identity is not present in MatNWB.

Do you think this should be fixed here, or in a separate branch? This problem is pre-existing, but now that HERD moved to hdmf_common, should we support it outright? or fine to add later?

bendichter (approved):

PR #506, opened 2023-05-17
Allow unit vectordata

lawrence-mbf:
Fixes #238 

## Motivation

See #238 and friends

## How to test the behavior?

See #238 and friends. A good example file is DANDIset 000022

- `waveform_mean`, `waveform_sd`, and `waveforms` attributes:
  - `unit`
  - `sampling_rate`
- `spike_times` attributes:
  - `resolution`

## Checklist

- [x] Have you ensured the PR description clearly describes the problem and solutions?
- [x] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/neurodatawithoutborders/matnwb/pulls) for the same change?
- [x] If this PR fixes an issue, is the first line of the PR description `fix #XX` where `XX` is the issue number?

PR #823, opened 2026-05-19
Breaking: remove legacy `VectorData` attribute patch

ehennestad:
## Summary

NWB schema 2.10 introduces `TimestampVectorData` and `DurationVectorData`, both subclasses of `VectorData` that define their own schema attributes (`unit`, `resolution`). MatNWB's legacy compatibility patch (PR #506) injected `unit`, `sampling_rate`, and `resolution` as hidden properties on the base `types.hdmf_common.VectorData` class. Because MATLAB does not allow a subclass to redeclare an inherited property with different attributes, the patched base class makes a valid `TimestampVectorData` impossible to generate under schema 2.10.

This PR removes the entire patch — all three hidden properties and the supporting codegen plumbing — so schema 2.10 can be generated cleanly.

## ⚠️ Breaking change

Setting `unit`, `resolution`, or `sampling_rate` directly on a `types.hdmf_common.VectorData` instance no longer works. Use the promoted properties on the parent `Units` container instead (available since #797).

```matlab
% Before
spikeTimes = types.hdmf_common.VectorData('data', t, 'description', '...');
spikeTimes.resolution = 1/20000;
units.spike_times = spikeTimes;

% After
units.spike_times = types.hdmf_common.VectorData('data', t, 'description', '...');
units.spike_times_resolution = 1/20000;
```

The same applies to `waveform_mean_sampling_rate`, `waveform_sd_sampling_rate`, `waveforms_sampling_rate`, `waveform_mean_unit`, `waveform_sd_unit`, and `waveforms_unit`. **Reading existing NWB files is unaffected** — the parser lifts these HDF5 attributes directly onto the parent `Units` container.

## What changed

**Codegen layer**
- Removed `patchVectorData` and its call site in `+file/processClass.m`.
- Removed the `fillVectorData*Conditional` helpers and the `preExportString` block in `+file/fillExport.m`.
- Removed `syncPromotedDatasetAttributeString` in `+file/fillSetters.m`.
- Removed the `promoted_to_container` property from `+file/Attribute.m`.
- Removed `+file/+internal/isPropertyHidden.m` and the `isHidden` mask in `+schemes/+internal/getRequiredPropsForClass.m`.
- Removed the `Hidden` / `Hidden, SetAccess = protected` property-block scaffolding in `+file/fillClass.m`.

**Regenerated types**
- `+types/+hdmf_common/VectorData.m`: no `unit` / `resolution` / `sampling_rate` properties, validators, constructor parsing, or conditional export branches.
- `+types/+core/Units.m`: each `postset_*_resolution` / `postset_*_sampling_rate` contains only the missing-dependency warning. No export-time backfill from child `VectorData` onto the parent.

**Tests**
- `+tests/+unit/+file/VectorDataPatchTest.m` asserts `unit`, `resolution`, and `sampling_rate` are absent from `types.hdmf_common.VectorData` and that `types.core.Units` does not sync `*_unit` from child `VectorData` instances.

## How to test the behavior?
```
runtests('tests.unit.file.VectorDataPatchTest')
runtests('tests.system.UnitTimesIOTest')   % PyNWB round-trip on the Units table
```

## Checklist

🤖 Generated with [Codex](https://chatgpt.com/codex)


ehennestad:
@bendichter 

Just realised that `TimestampVectorData` and `DurationVectorData` also defines the `resolution` attribute. 
In my opinion, the simplest fix is to remove the add-hoc `VectorData` patch entirely, as it was already made redundant in PR #797 

The drawback is that existing(?) code relying on setting the resolution of `VectorData` objects that go into the `waveform`* columns of a `Units` table will break, and it is not straight-forward to add a specific error message for such breakage. 

The property was already hidden and the syntax was not described in any tutorials, so I expect the number of affected users would be very small. 

Would appreciate your input!

bendichter (approved):

PR #830, opened 2026-06-18
Warn instead of error when reading values invalid per the NWB schema

### LINKED ISSUE #776: [Feature]: Warning instead of error when reading values that are not valid according to the NWB schema

When reading NWB files, different types of invalid values in the file will cause read errors. One example is if the schema specifies a attribute value to be constant (e.g. ElectricalSeries/data/unit should be **volts**), but the value of this attribute in the file is something else than the defined constant. For an example, see this discussion:
https://github.com/NeurodataWithoutBorders/helpdesk/discussions/75

Causing a file read error is perhaps too "dramatic", and a warning might be better.

@bendichter @rly @oruebel What are your thoughts here? 

Yes, files which have deviations from schema cannot be read. This can render a whole file useless, even if the issue is minor

I have two ideas in mind:

1. Very permissive read. Everything that is not according to schema causes a warning. The value being read is kept if possible.

2. A debug read mode. By default, reading files with invalid values causes an error, but support a debug flag / keyword argument that will change the error to a warning.

Yes.

### Code of Conduct

- [x] Have you ensured this change was not already [requested](https://github.com/NeurodataWithoutBorders/matnwb/issues)?

ehennestad:
Fix #776

⚠️ Depends on #844 and #845.

## Problem

MatNWB currently treats schema-invalid values as hard errors while reading NWB files. This is too strict: a single non-conforming value can make an otherwise usable file unreadable.

## Solution

Update validators to use `matnwb.common.validation.reportSchemaViolation`. Schema violations remain errors during normal object construction and export, but become warnings during file reads so the original values remain accessible.

The following validators are updated in this PR:

- Neurodata type validation: `checkType`
- Soft-link target validation: `validateSoftLink`
- Data type validation: `checkDtype`
- Constrained data type validation: `checkConstraint`
- DynamicTable consistency validation: `checkConfig` and `getOutermostIndexColumnName`

Constant-value validation (`checkConstant`) and shape validation (`validateShape`) are handled separately in #845.

## Tests

```matlab
nwbtest('Name', 'tests.unit.types.validators.CheckTypeTest')
nwbtest('Name', 'tests.unit.types.validators.ValidateSoftLinkTest')
nwbtest('Name', 'tests.unit.types.validators.CheckDtypeTest')
nwbtest('Name', 'tests.unit.types.validators.CheckConstraintTest')
nwbtest('Name', 'tests.unit.types.validators.DynamicTableCheckConfigTest')
nwbtest('Name', 'tests.unit.io.testCreateParsedType')
```

## Checklist

- [x] Have you checked to ensure that there are no other open or previously closed Pull Requests for the same change?
- [x] If this PR fixes an issue, is the first line of the PR description `Fix #XX` where `XX` is the issue number?

🤖 Generated with [Codex](https://chatgpt.com/codex)


bendichter:
decision here is to push out a fix for the problems users are facing today, specifically validation of dataset shape and schema-locked value (e.g. "volts"). Then we can perhaps think about expanding to other types of validation in the future.

PR #832, opened 2026-06-19
Validate property values on export to prevent writing invalid files

ehennestad:
⚠️ Depends on — merge these first:
 - #835
 - #836
 
### Background
MatNWB `v2.9.0` fixed some validation bugs rendering existing NWB files unreadable because of strict validation on read. In issue #776, the consensus is to relax validation on read, showing warnings instead of errors if validation fails.

PR #836 (merged) and PR #830 is ongoing work to support that.

### Problem
However, relaxing validation on read creates a new problem. MatNWB currently validates property values only on construction and modification, not on export. If users are allowed to read an NWB file with invalid values, they are also free to re-export those invalid values to new NWB files.

### Solution
This PR adds an export-time guard, ensuring that all properties are re-validated on file export.

## Changes

- Add `types.untyped.MetaClass/validateProperties`, invoked in `export` alongside the existing required-property and custom-constraint checks. For each set property that has a generated `validate_<prop>` method, it re-runs the validator and discards the result, so there are no setter side effects such as type coercion or column-name syncing. If the value is invalid it raises `NWB:Export:InvalidPropertyValue`, including the property name and file location, and preserves the original validator error as a cause.
- Empty/unset optional properties are skipped, matching what `export` writes.
- Validators run in the default (strict) context here, so a schema violation is an error on export even when the same check is lenient during a read.

## Relationship to other work

This is the write-side counterpart to the read-side permissiveness in #776 — read lenient (warn, keep value), write strict (error). It generalizes the pattern already used for `DynamicTable` (`checkConfig` via the custom-constraint path) to all property validators. Its user-visible effect lands together with #776, so this is best reviewed and merged alongside that work.

## How to test

`tests.unit.types.MetaClassValidatePropertiesTest` exercises the guard with a small `MetaClass` test double: an invalid property value raises `NWB:Export:InvalidPropertyValue` with the property name, location, and original cause; an unset (empty) property is skipped; a valid value passes.

## Checklist

- [x] Have you checked to ensure that there aren't other open or previously closed Pull Requests for the same change?


bendichter:
I don't understand the type of invalid workflow you are trying to prevent

bendichter (approved):

PR #835, opened 2026-06-24
fix: export scalar datetime as a scalar dataset

ehennestad:
### Problem

When a scalar `datetime` property such as `session_start_time` is set, `types.util.correctType` ran `num2cell` on it, converting the scalar `datetime` into a 1×1 cell array before formatting. On export that 1×1 cell was written to HDF5 as a 1-element, *non-scalar* dataset instead of a scalar dataset.

So a value that is conceptually scalar was stored with the wrong dataspace. The exported file no longer reflected the field's scalar shape, the value did not round-trip faithfully (a scalar went in, a 1-element array came back), and on read it was returned as a lazy `DataStub` instead of an eagerly-loaded scalar.

### Solution

Drop the `num2cell` conversion and format datetimes in place, so the `datetime` branch of `correctType` preserves the class and shape of its input:

- a scalar `datetime` stays a scalar `datetime`
- a `datetime` array stays a `datetime` array (formatted element-wise)
- a cell array of datetimes stays a cell array of datetimes

Scalar datetime properties are now written as scalar datasets. Covered by a test asserting the scalar dataspace on export, plus an updated reader test reflecting that a scalar dataset is read eagerly rather than as a `DataStub`.

## How to test the behavior?

```matlab
% session_start_time is a scalar datetime
nwb = tests.factory.NWBFile();
nwbExport(nwb, "scalar_datetime.nwb");

info = h5info("scalar_datetime.nwb", "/session_start_time");
disp(info.Dataspace.Type)   % "scalar"  (previously a 1-element, non-scalar dataset)
```

## Checklist


bendichter:
`NWBFile.session_start_time` should be a scalar. However, `NWBFile.file_create_date` should be a 1x1 dataset. Even though it is only a single element when it is first written, it is appended every time the file is edited. Does this PR allow for that?

https://nwb-schema.readthedocs.io/en/latest/format.html#nwbfile

bendichter:
Please add a test for ensuring that file_create_date continues to be a dataset

ehennestad:
> Please add a test for ensuring that file_create_date continues to be a dataset

@bendichter Added that test here: https://github.com/NeurodataWithoutBorders/matnwb/pull/835/commits/8f162183ea1ae71beec8b3130684de0e9c033f40

bendichter (approved):

PR #860, opened 2026-08-21
Add support for persistent object ids

ehennestad:
## Motivation

**Problem** — Every NWB object is supposed to carry a stable `object_id` (UUID), which downstream tools and the HERD external-resources tables use to reference objects. MatNWB wrote this attribute but regenerated it on every export and discarded it on read: re-saving a file silently re-identified every object, a round trip (`nwbRead` → `nwbExport`) produced a file whose objects no longer matched the original, and the id was not accessible from MATLAB at all. This makes HERD unusable in practice — any id a user puts in a HERD table is invalidated by the next export — and breaks any workflow that treats object ids as stable keys.

**Solution** — Object ids are now persistent, matching PyNWB/HDMF semantics: an id is assigned once when an object is constructed, adopted from the file on read, and reused verbatim on every export. This is groundwork for HERD support, where objects are referenced by their `object_id`. For workflows that intentionally fork a file into a new entity, a new `generateNewObjectId` method assigns fresh ids on demand.

## What changed

- Every neurodata type object has a read-only `object_id` property, assigned at construction.
- `nwbRead` adopts the ids stored in the file, including for typed datasets such as `DynamicTable` columns.
- `nwbExport` reuses the stored id, so ids are stable across repeated exports, in-place edits, and read → export round trips.
- Files without `object_id` attributes (written before NWB 2.2 or by older tools) read cleanly; their objects receive newly generated ids.
- New method `generateNewObjectId()` assigns a fresh id to an object and, by default, recursively to all typed objects contained in it (`Recurse=false` limits it to the object itself), mirroring HDMF's `Container.generate_new_id`.
- Exporting one object to two locations in the same file now warns (`NWB:Export:DuplicateObjectId`). Persistent ids make this visible: the two locations would otherwise share a single id, which PyNWB rejects when reading the file. The second location is written as a copy under a newly generated id — the layout MatNWB has always produced — so existing code keeps working. A future release can turn this into an error.
- The `intro` and `ecephys` tutorials no longer write the same object to two locations. `ecephys` now links its shared electrode table region with a `types.untyped.SoftLink` instead of writing five copies of it.
- Internal cleanup: the parse-time skip list previously hardcoded as a "blacklist" struct in `nwbRead` and `io.parseGroup` is centralized in `io.internal.defaultParseExclusions`.

- `object_id` is a hidden, `SetAccess = private` property on `types.untyped.MetaClass`. The `MetaClass` constructor adopts an id passed as a name-value argument (the read path) or generates a new UUID; all generated type constructors already forward their arguments up the superclass chain, so no generated code changed.
- `io.parseDataset` previously routed only attributes matching public properties into a typed dataset's constructor; `object_id` is now explicitly included so it is consumed by the typed object instead of being promoted to the parent under an elided name (e.g. `id_object_id`).
- The `NwbFile` constructor's arguments block accepts `object_id` explicitly, analogous to `nwb_version`.
- The writer records the location each object id is written to, which is how a second location is detected. Re-registering the same location is allowed, because reference resolution re-exports an object at its original location.

</details>

## Examples

Shared setup:

```matlab
nwb = NwbFile('identifier', 'demo', 'session_description', 'demo', ...
    'session_start_time', datetime(2026,8,21,12,0,0, 'TimeZone', 'local'));
```

### Object ids persist across exports and round trips

```matlab
nwbExport(nwb, 'demo.nwb');
fprintf('object_id after 1st export: %s\n', h5readatt('demo.nwb', '/', 'object_id'));
nwbExport(nwb, 'demo.nwb');
fprintf('object_id after 2nd export: %s\n', h5readatt('demo.nwb', '/', 'object_id'));

nwbIn = nwbRead('demo.nwb', 'ignorecache');
nwbExport(nwbIn, 'copy.nwb');
fprintf('  original: %s\n', h5readatt('demo.nwb', '/', 'object_id'));
fprintf('  copy:     %s\n', h5readatt('copy.nwb', '/', 'object_id'));
```

**Before** — a new id on every export; the copy no longer matches the original:

```
object_id after 1st export: 1a238a76-f1bd-4534-9800-91ae4374ac65
object_id after 2nd export: 25313b6a-e01b-4de4-a5d3-1f8117d13bb0
  original: 25313b6a-e01b-4de4-a5d3-1f8117d13bb0
  copy:     e1b33688-ce7c-4f50-807c-519c7efe9359
```

**After** — one id, everywhere:

```
object_id after 1st export: 9fc466af-03f2-4042-b706-b7f56a156815
object_id after 2nd export: 9fc466af-03f2-4042-b706-b7f56a156815
  original: 9fc466af-03f2-4042-b706-b7f56a156815
  copy:     9fc466af-03f2-4042-b706-b7f56a156815
```

### Accessing and regenerating ids from MATLAB

```matlab
nwb.object_id
```

**Before**:

```
Error: Unrecognized method, property, or field 'object_id' for class 'NwbFile'.
```

**After**:

```
nwb.object_id --> 9fc466af-03f2-4042-b706-b7f56a156815
```

Intentionally forking a file as a new entity:

```matlab
nwbIn.generateNewObjectId();   % new ids for the file object and everything in it
```

```
after generateNewObjectId --> 1bee88c6-d194-4688-80f2-46fd55a10c18
```

### Exporting one object to two locations

```matlab
trials = types.core.TimeIntervals('colnames', {'start_time','stop_time'}, ...
    'description', 'trial data and properties');
trials.addRow('start_time', 0.1, 'stop_time', 1.0);

nwb.intervals_trials = trials;
nwb.intervals.set('custom_intervals_table_name', trials);   % the same object again
nwbExport(nwb, 'demo.nwb');
```

**Before** — silently wrote two copies of the table, each with its own freshly generated id.

**After** — the copy is still written, so the file stays readable, but the duplication is now reported. One warning is raised per duplicated object, i.e. for the table and for each of its columns:

```
[Warning: The object of type "types.core.TimeIntervals" was already exported to
the file location "/intervals/custom_intervals_table_name" and is now also
exported to "/intervals/trials". Each neurodata object should have exactly one
location in an NWB file, so "/intervals/trials" is written as a copy under a
new object id. Create a separate object for each location, or use a
types.untyped.SoftLink to reference the object at
"/intervals/custom_intervals_table_name".]
```

```
ids written:
  /intervals/trials                      b49fc061-d6a3-4b96-a0aa-646b10dd302f
  /intervals/custom_intervals_table_name 48c9ddcd-f1aa-4746-9f03-b986702c0baa
```

To store a single object referenced from both places, link it instead — this is what the `ecephys` tutorial now does for its shared electrode table region:

```matlab
'electrodes', types.untyped.SoftLink(electrode_table_region), ...
```

## How to test

```matlab
nwb = NwbFile('identifier', 'demo', 'session_description', 'demo', ...
    'session_start_time', datetime(2026,8,21,12,0,0, 'TimeZone', 'local'));
nwbExport(nwb, 'demo.nwb');
nwbExport(nwb, 'demo.nwb');
assert(strcmp(h5readatt('demo.nwb', '/', 'object_id'), nwb.object_id))

nwbIn = nwbRead('demo.nwb', 'ignorecache');
nwbExport(nwbIn, 'copy.nwb');
assert(strcmp(h5readatt('copy.nwb', '/', 'object_id'), nwb.object_id))
```

Unit tests covering construction, round trips, typed dataset children, legacy files without ids, `generateNewObjectId`, and the duplicate-location warning are added in `+tests/+unit/objectIdTest.m`. They build their neurodata types through two new shared helpers, `tests.factory.TimeSeries` and `tests.factory.TimeIntervals`, and the base writer smoke test covers the writer's new method.

## Checklist


bendichter:
> Exporting one object to two locations in the same file now warns (NWB:Export:DuplicateObjectId). Persistent ids make this visible: the two locations would otherwise share a single id, which PyNWB rejects when reading the file. The second location is written as a copy under a newly generated id — the layout MatNWB has always produced — so existing code keeps working. A future release can turn this into an error.

Does this mean that we are knowingly creating NWB files that PyNWB cannot read?

ehennestad:
> the two locations would otherwise share a single id

No. 

Prior to this PR, for the `intro` and `ecephys` tutorial, the same neurodata object was written to file multiple times (in different locations). And before this PR, each time an object was exported, it got a new UUID, and so there was no problem reading it back with PyNWB.

Now, however, each object has a persistent UUID, and exporting that object multiple times would produce files that would fail to read with PyNWB. 

The solution is to show a warning when the same object is exported in multiple locations, but then "mint" a new uuid for each export location of a single object (this emulates pre-PR behaviour). 

The behaviour in PyNWB is to silently create softlinks when one object is exported to multiple locations. I did not want to do that here, because 1) Creating softlinks would silently change the structure of the file and 2) MatNWB does not resolve softlinks the same way as PyNWB on read, so users would end up with files that looked different than before if they re-use older conversion scripts.

bendichter:
ok got it, thanks for the explanation

bendichter (approved):

PR #884, opened 2026-08-27
ci: fix and harden the prepare release workflow

ehennestad:
## Motivation

**Problem** — The "Prepare release" workflow is still blocked and can not produce a release. Badge generation fails with `ModuleNotFoundError: No module named 'pkg_resources'`.

Nothing in this repository changed. pybadges imports `pkg_resources`, which ships as part of setuptools, and the workflow installs the newest setuptools on every run. setuptools removed `pkg_resources` in 82.0, released 2026-02-08, so every run since then has installed a setuptools that cannot satisfy pybadges. The last successful release, v2.10.0, ran before that date. pybadges was archived in 2024, so there will be no fix upstream.

The same run exposed a second problem. Tagging ran under `if: always()`, so the workflow pushed the `v2.11.0` tag even though it had generated no badges and created no release. That tag then blocks a retry, because the next run stops at `git tag -a`.

**Solution** — Generate the badges with anybadge, which is maintained and needs neither setuptools nor `pkg_resources`. Tag only once the steps before it have succeeded, and let a retry reuse a tag that already points at the commit being released.

## Also in this PR

Reviewing the rest of the workflow turned up further ways it can go wrong. None of them caused this failure, and all are cheap to close while the file is open.

**Dispatching from a branch now rehearses instead of publishing.** The test matrix checked out the dispatched ref while the release job hardcoded `main`, so a dispatch from a feature branch tested that branch and then released `main` — a green matrix certifying code that did not ship. The release job now checks out the ref it was dispatched from, and only publishes when that ref is `main`. Dispatching from a branch still runs every step, including the version rewrite and badge generation, so it remains a useful way to check changes to this workflow before they land; it just pushes nothing and creates no release. The run summary says which mode it used.

**An already released version is rejected before the test matrix runs.** Previously the version bump reached `main` first and only the tag step objected, leaving a stray commit on a branch that takes the deploy key to correct. A tag on the current tip of `main` is still accepted, so a run that failed after tagging can still be retried.

**A changed `Contents.m` separator now fails the job instead of truncating the file.** The header is rewritten from content extracted after the dashed separator line. If that line stops matching, the extraction returns nothing and the file is replaced by its header alone, then pushed to `main` with no PR to catch it.

**Runs for a given ref are serialised**, so two dispatches cannot race on the push to `main`, the tag and the badge branch.

**The release and JUnit jobs state the permissions they need**, so both keep working if the repository default for `GITHUB_TOKEN` is ever reduced to read-only.

**The version input reaches shell steps as an environment variable** rather than being pasted in with `${{ }}`. Interpolation happens before the script runs, so a version containing a quote was parsed as code, and the format check could not prevent it because that check ran afterwards. Dispatching is limited to users with write access, so this was hardening rather than an open door. The release list for the tested-with badge moved for the same reason: it was interpolated into a Python string literal, where an apostrophe would have been a syntax error.

**anybadge is pinned**, so a new release cannot alter the badges or break the job until the pin is updated deliberately.

**The `latest/` badge is only rewritten when the version being released is newer than every existing tag.** It was rewritten on every run, so rebuilding an older release pointed `latest` at that older schema range.

## How to test

This branch can be checked with the feature it adds. Dispatch "Prepare release" from `claude/draft-release-job-issues-46f330` with a version number: the job runs end to end against this branch and publishes nothing.

The dependency failure can also be reproduced directly, in the environment the runner uses:

```bash
python3 -m venv /tmp/badge-check
/tmp/badge-check/bin/pip install --upgrade setuptools
/tmp/badge-check/bin/python -c "import pkg_resources"          # fails, as on the runner

/tmp/badge-check/bin/pip install pybadges
/tmp/badge-check/bin/python -c "from pybadges import badge"     # fails: the current workflow

/tmp/badge-check/bin/pip install anybadge
/tmp/badge-check/bin/python .github/workflows/scripts/create_nwb_schema_version_badge.py 2.11.0
```

The stale `v2.11.0` tag from the failed run has been deleted, so a release dispatch from `main` after this merges starts clean.

## Checklist

### DIFF
--- a/.github/workflows/prepare_release.yml
+++ b/.github/workflows/prepare_release.yml
@@ -9,6 +9,12 @@ on:
         required: true
         type: string
 
+# Releases from the same ref run one at a time. Rehearsals on a branch use a
+# separate group, so they never queue behind or block a release from main.
+concurrency:
+  group: prepare-release-${{ github.ref }}
+  cancel-in-progress: false
+
 jobs:
   validate_version:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


ehennestad:
"Rehearsal" run passes with changes from this PR:
https://github.com/NeurodataWithoutBorders/matnwb/actions/runs/33112560720

bendichter (approved):

PR #885, opened 2026-08-27
docs(tutorials): add a tutorial for external resource references

ehennestad:
## Motivation

Add a tutorial for the HERD API, mirroring the [PyNWB external resources tutorial](https://pynwb.readthedocs.io/en/latest/tutorials/general/plot_external_resources.html) so a reader can move between the two APIs.

**Preview tutorial on ReadTheDocs:** 
https://matnwb--885.org.readthedocs.build/en/885/pages/tutorials/externalResources.html

## What changed

- `tutorials/externalResources.mlx` records and reads back external resource references on a file it builds itself: a subject's species, and a column of an electrodes table.
- The tutorial has a page under General Tutorials.

Storing a HERD outside a file is left out, because `to_zip` and `from_zip` are not implemented in MatNWB.

## How to test

```matlab
nwbtest('Name', 'tests.system.tutorial.TutorialTest')
```

## Todo
- [x] Verify and approve PRs below in stack
- [x] Verify tutorial is clear to follow and covers relevant parts of API

## Checklist


ehennestad:
> Pass the table as the object and the column name as Attribute. The reference is attached to the column itself, which is the object that holds the values, rather than to the table. A ragged column works the same way: Attribute names the column holding the values, and the index that groups them into rows is a separate property.

The last half of this paragraph is unclear - Either improve or drop it completely (The main point here is that the column itself is added as an attribute, not an index column)

bendichter (approved):

PR #888, opened 2026-09-22
feat(types): add TimeSeries.getDataInUnits and applyConversion

### LINKED ISSUE #429: [Feature]: Autoload TimeSeries data with offset, conversion and channel_conversion

Hi,

Instances of `TimeSeries` and children classes have additional fields added that specify how to properly load data. It would be useful if the load function(s) can automatically apply the operations expected to be performed.

_No response_

Here's an example of how this could work for backwards compatibility. If any new fields are added in the future or to a child class of `TimeSeries` then they can just be stacked in the `cfg` variable and placed in the correct place of the operation on the bottom.

```
fname = '/Users/sportsnoah14/Documents/Northshore_Hosptial/HBML/PROJECTS/Movies/sub-NS174_ses-implant02_task-DespMeEng_ieeg.nwb';
nwbfile = nwbRead('myfile.nwb');

ts = nwbfile.resolve('/acquisition/test_ts');
p = properties(ts);
data = ts.data.load;

cfg = [];
cfg.channel_conversion = ones(size(data,1), 1);;
cfg.offset = 0;
cfg.data_conversion = 1;
operation_fields = fieldnames(cfg);
for ii = 1:length(operation_fields)
    op = operation_fields{ii};
    if any( strcmp(op, p) )
        if ~isempty( ts.(op) )
            cfg.(op) = ts.(op);
        end
    end
end

data = data*cfg.data_conversion.*cfg.channel_conversion + cfg.offset;
```

Yes, but I would need guidance.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/neurodatawithoutborders/matnwb/blob/master/.github/CODE_OF_CONDUCT.md)

ehennestad:
fix #429

## Motivation

**Problem** — The values stored in a `TimeSeries` are usually not in the unit its `data_unit` property names. Reaching that unit means applying `data_conversion` and `data_offset`, and for an `ElectricalSeries` the per-channel `channel_conversion` as well. Doing it by hand has three traps:

1. Not knowing there is anything to do. `data_unit` names the unit the values are meant to be read in, not the unit they are stored in, and nothing in the array itself says which, so a reader who does not know the format reads the stored values as if they were already in that unit.
2. Scaling the stored integers directly keeps the result an integer, so `int16` counts scaled by a conversion factor come back rounded and saturated instead of in volts.
3. MatNWB reverses the dimension order of a dataset on read, so the channel factors have to be applied along the second-to-last dimension of the array (the first for `[channels x time]` data) even though `channel_conversion/axis` names axis 1 of the dataset as it is laid out in the file.

PyNWB exposes this as `TimeSeries.get_data_in_units`; MatNWB had no equivalent.

**Solution** — Add two methods on a handwritten `matnwb.neurodata.TimeSeriesBase`, attached to the generated `types.core.TimeSeries` through the `customBaseClasses` hook in `file.fillClass` that `DynamicTable` already uses. `getDataInUnits` scales the whole dataset. `applyConversion` scales data the caller has already loaded, so a subset of a dataset too large to hold in memory can be scaled the same way.

## What changed

- `getDataInUnits()` returns `data * data_conversion + data_offset`, with `channel_conversion` applied per channel when the type defines it. It reads the whole dataset into memory.
- `applyConversion(rawData)` applies the same scaling to data already in memory.
- `applyConversion(rawData, 'Channels', channels)` names the channels a partial load holds, which is what a per-channel `channel_conversion` needs in order to pick the right factors.
- Raw data is promoted to a floating point type before scaling, so integer data does not round or saturate. The result is `double`, or `single` when the stored data is `single`.
- `data` held as a `DataStub`, `DataPipe` or `ExternalLink` is read into memory by `getDataInUnits`.
- `types.core.TimeSeries` is regenerated with `generateCore('2.10.0')` after the `fillClass` change, rather than edited by hand.
- A how-to page, *Reading TimeSeries Data in Units*, under *Work with Data* in the docs. Handwritten methods are not otherwise listed on the site, so without it the two methods are only discoverable through `methods(timeSeries)`.

Two things differ from the PyNWB implementation because of MATLAB and HDF5 semantics.

PyNWB multiplies `self.conversion * self.channel_conversion` and lets numpy broadcast, which works because the channel axis is last in the file's C ordering. MatNWB reads the dataset with its dimensions reversed, so the channel factors are reshaped to broadcast along dimension `rank - axis`. The rank is that of the stored `data` property, not of the array passed in, because MATLAB drops trailing singleton dimensions from a slice and a time slice of 3-D data would otherwise put the channels on the wrong dimension. The zero-based `channel_conversion_axis` is used when the type carries it.

MATLAB integer arithmetic rounds and saturates rather than promoting, so the raw data is cast before the conversion factor is applied. Without that cast the first example below returns `[-3; 0; 2]` instead of volts.

The channel count is read from the shape of the `data` property itself — `dims` on a `DataStub`, `size()` on a `DataPipe` or an in-memory array — without loading it. A `channel_conversion` whose length differs from that count raises `ChannelConversionMismatch` whatever subset was passed, so an inconsistent file is reported the same way from every path. Export validates `channel_conversion` and `data` independently, so such a file can be written. Once they agree, a `rawData` with fewer channels than the TimeSeries and no `'Channels'` raises `ChannelSelectionRequired` rather than guessing which channels were loaded, and one with more raises `DataMismatch`, since it cannot have come from this TimeSeries.

`channel_conversion` and `channel_conversion_axis` are declared by subclasses such as `ElectricalSeries`, not by `TimeSeries`, so `TimeSeriesBase` reaches them through `isprop` and dynamic property access.

The validation failures are thrown as the caller, so the console header names `applyConversion` or `getDataInUnits` rather than the private helper the checks live in. A fault in the reshaping itself still reports against the frame it happens in.

</details>

### Scaling a whole dataset

```matlab
timeSeries = types.core.TimeSeries( ...
    'data', int16([-32768; 0; 32767]), ...
    'data_unit', 'volts', ...
    'data_conversion', 2.5/32768, ...
    'starting_time', 0, 'starting_time_rate', 30000);
```

Scaling the stored counts directly keeps the result an `int16`:

```
>> timeSeries.data * timeSeries.data_conversion

   -3
    0
    2
```

```
>> timeSeries.getDataInUnits()

   -2.5000
         0
    2.4999
```

### A per-channel conversion

```matlab
electricalSeries = types.core.ElectricalSeries( ...
    'data', ones(4, 3), ...
    'data_conversion', 10, ...
    'data_offset', 1, ...
    'channel_conversion', [0.5; 1; 2; 4]);
```

```
>> electricalSeries.getDataInUnits()

     6     6     6
    11    11    11
    21    21    21
    41    41    41
```

### Scaling a subset without reading the whole dataset

```
>> rawData = electricalSeries.data([2 4], :);
>> electricalSeries.applyConversion(rawData, 'Channels', [2 4])

    11    11    11
    41    41    41
```

Omitting `'Channels'` for a partial load is refused rather than guessed:

```
>> electricalSeries.applyConversion(rawData)
Error using matnwb.neurodata.TimeSeriesBase/applyConversion (line 141)
Dimension 1 of data holds 2 of the 4 channels of this TimeSeries. Name the channels it holds
with the "Channels" argument, for example applyConversion(data, 'Channels', 1:2).
```

## How to test

```matlab
nwbtest('Name', 'tests.unit.types.TimeSeriesTest*')
```

## Checklist


bendichter (approved):

ehennestad:
Two changes landed after the approval.

- e08752ba fixes how the channel dimension is found. It was taken from `ndims` of the array passed to `applyConversion`; MATLAB drops trailing singleton dimensions, so a time slice `data(:, :, k)` of 3-D `ElectricalSeries` data arrived as 2-D and its channels were looked for on dimension 1. The dimension now comes from the rank of the stored `data` property. Covered by `testApplyConversionOnTimeSliceOfThreeDimensionalData`.
- 42439636 consolidates `TimeSeriesTest` from 30 tests to 17. Each behavior is covered once; nothing that was tested is dropped.

The remaining commits are doc edits to the how-to page and the help of `getChannelDimension`.

PR #890, opened 2026-09-23
feat(herd): reference plain-value attributes with a relative path

ehennestad:
## Motivation

**Problem** — HDMF's `add_ref` can attach an external reference to an attribute that is not a neurodata type, such as `Subject.species` or the unit of a `TimeSeries`. The objects table then records the container together with the attribute's schema path below it as `relative_path` (`TimeSeries.unit` becomes `data/unit`). #877 left this out: `addRef` with such an attribute raised `NWB:HERD:UnsupportedAttribute`, and MatNWB always wrote `relative_path` empty. The obstacle was that MatNWB flattens nested schema nodes into underscore-joined property names (`data/unit` → `data_unit`), and the name alone does not say where the underscores were slashes.

**Solution** — The code generator already knows each property's schema path: `fillExport` uses it to write the export paths. `fillClass` now uses the same resolution to list, per generated class, the schema paths of its plain-value properties in a private `SchemaRelativePaths` constant, following the pattern of `DeclaredSchemaColumns`. At add time, `HERDBase.resolveTarget` flattens each listed path with `_` and records the path whose flattened form names the requested property. Going from path to name is unambiguous, so no runtime spec walking is needed.

## What changed

- `nwb.addRef(container, Attribute=name, ...)` accepts a property holding a plain value. The reference is recorded on `container` with the property's schema path as `relative_path`, for example `species`, `data/unit` or `general/experimenter`.
- A property holding a neurodata type is still referenced as that object with an empty `relative_path`, as before.
- Links, sets of neurodata types (such as `acquisition`) and values nested in a typed node still raise `NWB:HERD:UnsupportedAttribute`. The error message now also covers classes generated by an earlier MatNWB, which have no `SchemaRelativePaths` and need regenerating.
- `file.internal.traverseRaw` is moved out of `fillExport` so both `fillExport` and `fillClass` resolve paths with it. Regenerated export methods are byte-identical.
- Generated classes in `+types` are regenerated. Each changed class gains only the `SchemaRelativePaths` constant; classes without plain-value properties of their own are unchanged.
- Help text of `addRef`, `NwbFile.addRef` and `getObjectType` describes the new behavior.

HDMF resolves the path in `HERD._resolve_object_target` (`_check_object_field` before 6.0): it takes the attribute's spec from the container's ObjectMapper and strips the first segment of `spec.path`. The rule is the same from hdmf 3.14 through 6.2.

The listed paths were compared against HDMF's ObjectMapper for all 95 core and hdmf-common types of schema 2.10.0 and match exactly. The comparison found one case the generator initially got wrong: `Units.spike_times_resolution` and the `waveform*` attributes sit on `spike_times` and `waveforms`, which are typed `VectorData` datasets declared inside `Units`. HDMF does not list them among `Units`' attributes, because their nearest neurodata type is the `VectorData`. Paths that pass through a typed node are therefore not listed.

Each class lists only the paths of properties it declares itself, and `collectConstantPropertiesAcrossHierarchy` adds those of its parent types, as for `DeclaredSchemaColumns`. A keyword-remapped property name (`events_`) is mapped back to its schema name through `SchemaPropertyNameMapping` before the comparison.

</details>

## Example

```matlab
nwb.addRef(nwb.general_subject, Attribute="species", Key="Mus musculus", ...
    EntityId="NCBITaxon:10090", EntityUri="http://purl.obolibrary.org/obo/NCBITaxon_10090");
nwb.addRef(timeSeries, Attribute="data_unit", Key="meters", ...
    EntityId="UO:0000008", EntityUri="http://purl.obolibrary.org/obo/UO_0000008");
```
```
>> references = nwb.getExternalResources().toTable();
>> references(:, ["object_type", "relative_path", "key", "entity_id"])

  2×4 table

     object_type      relative_path          key                entity_id     
    ______________    _____________    ________________    ___________________

    {'Subject'   }    {'species'  }    {'Mus musculus'}    {'NCBITaxon:10090'}
    {'TimeSeries'}    {'data/unit'}    {'meters'      }    {'UO:0000008'     }
```

## How to test

```matlab
nwbtest('Name', 'tests.unit.HERDTest*')
nwbtest('Name', 'tests.unit.file.SchemaRelativePathsTest*')
```

`HERDTest/testPynwbReadsHerd` checks that pynwb finds a reference MatNWB wrote under `relative_path="species"`.

## Checklist


ehennestad:
@bendichter @rly 

```matlab
% Create a file
nwb = tests.factory.NwbFile()

% Add a subject
nwb.general_subject = types.core.Subject( ...
    'subject_id', '001', ...
    'description', 'a mouse used in this session', ...
    'species', 'Mus musculus', ...
    'sex', 'M', ...
    'age', 'P90D');

% Add a reference
nwb.addRef(nwb.general_subject, ...
    Attribute = "species", ...
    Key = nwb.general_subject.species, ...
    EntityId = "NCBITaxon:10090", ...
    EntityUri = "http://purl.obolibrary.org/obo/NCBITaxon_10090");
```

**Get references for type, filtered by attribute:**
```Matlab
herd = nwb.getExternalResources()
herd.getObjectType("Subject", RelativePath="species")
```

**Output:**
```text
ans =

  1×8 table

                 file_object_id                                object_id                    object_type    relative_path      field             key                entity_id                             entity_uri                    
    ________________________________________    ________________________________________    ___________    _____________    __________    ________________    ___________________    __________________________________________________

    {'9534af9a-a03d-46e8-b575-03f01cb2e99b'}    {'e2519b5f-743c-4036-8a11-8a6a592097fc'}    {'Subject'}     {'species'}     {0×0 char}    {'Mus musculus'}    {'NCBITaxon:10090'}    {'http://purl.obolibrary.org/obo/NCBITaxon_10090'}
```

Returns one reference, as expected.

**Get references for type, no attribute filter:**
```Matlab
herd.getObjectType("Subject")
```

**Output:**
```text
ans =

  0×8 empty table

    file_object_id    object_id    object_type    relative_path    field    key    entity_id    entity_uri
    ______________    _________    ___________    _____________    _____    ___    _________    __________
```

Returns zero references. I would have expected to see all references for a subject and its attributes. (the current implementation is mirroring hdmf

